PR #38, opened 2021-08-26
Enable cross validation support

wiseodd:
@runame please review


runame:
The change to support regression looks good. There might be loss functions which also require the output variance, but for now the mean is enough. 

Regarding the example script, I think it is very similar to the existing regression example, so maybe we can create a new directory `examples` and collect examples which we don't explicitly explain in the docs there. We could also do this on a separate branch for now. What do you think?

runame (commented):
The change to support regression looks good. There might be loss functions which also require the output variance, but for now the mean is enough.

Regarding the example script, I think it is very similar to the existing regression example, so maybe we can create a new directory `examples` and collect examples which we don't explicitly explain in the docs there. We could also do this on a separate branch for now. What do you think?

wiseodd:
Updated! 

* `utils.py` always return variance whenever available
* `optimize_prior_precision` has an optional argument `loss_with_var` to handle your concern
* added `example` directory

runame:
@wiseodd Is it ok if I remove the `examples/regression_example_mc.py` file? If I remember correctly we decided to only have examples in this repository which will also be shown in the docs, i.e. the original regression one, your new calibration example, and then the continual learning one.

wiseodd:
Yes, please go ahead

PR #41, opened 2021-10-12
Remove abc

metodj:
- removed all ABC stuff (abstract methods etc.)
- refactored baselaplace.py -> BaseLaplace now has two subclasses, ParametricLaplace and FunctionalLaplace (this will greatly help with bringing in the GP inference)


metodj:
Also, in the `laplace.py` I changed it so that only subclasses of `ParametricLaplace` are exposed through `Laplace` for the time being. Once `FunctionalLaplace` is ready, we can switch back to `BaseLaplace`. Let me know if that's fine.

AlexImmer:
Okay, we can leave the plain `NotImplementedError`s then. That's actually verbose enough. I think you only need to recompile the documentation (see the readme for this) and then push the changed documents. Let me know if that doesn't work for you. 
@runame Are you okay with merging this in?

metodj:
Thanks for pointing out the convention for the string style, wasn't aware of it before

Issue #44, opened 2021-10-19
Integration of `asdfghjkl-0.1`


aleximmer:
Current version can be found [here](https://github.com/kazukiosawa/asdfghjkl/tree/0.1). For example, Kazuki Osawa mentioned that the `data_average` parameter now defaults to `True` but we require `False` for a proper Hessian approximation.

AlexImmer:
Kazuki postponed the release of the next version of his library to end of January/beginning of February. This issue is therefore removed from NeurIPS pre-release.

wiseodd:
This is part of #144. The `mc-subset2` of that PR was a branch of `integrate-asdl-latest`. Once it's merged, this should be closed.

PR #47, opened 2021-10-27
[WIP] Integration of the latest `asdl` package

aleximmer:
Enables regression and provides new interface for several methods that need to be synced (corresponds to #44). 
There are two key quantities that require more investigation and tests:
- [ ] regression empirical fisher seems to mismatch our backpack variant (currently failing test)
- [ ] regression fisher with sampling (`stochastic=True`) currently untested


runame:
When this PR is ready and our backend also supports regression with ASDL, we can also change the default backend to `AsdlGGN`. This avoids issues with quite a few architectures, which are not implemented via a `Squential`, e.g. #92.

adamxyang:
Hi, I'm also interested in the integration with the new ASDL. Does the current branch with the new ASDL work with classification tasks?

adamxyang:
Looks like changing `fisher_maker = get_fisher_maker(self.model, cfg, self.kfac_conv)` to `fisher_maker = get_fisher_maker(self.model, cfg)` in `asdl.py` line 160 makes things work with the classification example.

wiseodd:
Yes, that hotfix should work without any side effects. Let us know if you found any bugs, though!

aleximmer:
@wiseodd you built on these changes for #144, right? Then, let's close this PR. Do you know if the two issues I report above are resolved?

wiseodd:
Yes, should be good. Both are already covered with passing tests in #144. Closing this.

Issue #49, opened 2021-11-05
Example: post-hoc predictive uncertainty 


AlexImmer:
Something like CIFAR-10 pretrained model loading and show how calibration improves.

PR #54, opened 2021-11-11
Calibration example

wiseodd:
I put the example in a dedicated directory `./examples` since otherwise it will be too untidy at the root directory. Please let me know your thoughts.


wiseodd:
This is for addressing #49

examples/calibration_example.md
```diff
@@ -0,0 +1,111 @@
...
+
+``` python
+model = wrn.WideResNet(16, 4, num_classes=10).cuda().eval()
+
+# Download pre-trained model if necessary
+if not os.path.isfile('CIFAR10_plain.pt'):
+    if not os.path.exists('./temp'):
+        os.makedirs('./temp')
+
+    urllib.request.urlretrieve('https://nc.mlcloud.uni-tuebingen.de/index.php/s/2PBDYDsiotN76mq/download', './temp/CIFAR10_plain.pt')
```
AlexImmer, line 43: Maybe these 4 lines can be hidden away in the `helper` methods and wrapped in `download_model_temp()` or something alike?

examples/calibration_example.md
```diff
@@ -0,0 +1,111 @@
...
+We are now ready to see how calibrated is the model. The metric we use is the expected calibration error (ECE, Naeni et al., AAAI 2015). Note that lower values are better for this metric.
+
+First, let us inspect the MAP model. We shall use the [`netcal`](https://github.com/fabiankueppers/calibration-framework) library to easily compute the ECE.
+
+``` python
+probs_map = predict(test_loader, model, laplace=False)
+acc_map = (probs_map.argmax(-1) == targets).mean()
+ece_map = ECE(bins=15).measure(probs_map, targets)
+
+print(f'[MAP] Acc.(🠕): {acc_map:.1%}; ECE(🠗): {ece_map:.1%}')
```
AlexImmer, line 75: What is the meaning of "🠕" in the brackets? Is it some unicode thing that doesn't render properly?

examples/helper/dataloaders.py
```diff
@@ -0,0 +1,131 @@
+import torch
```
AlexImmer, line 1: Why are all these data sets available if we only need one?

examples/helper/wideresnet.py
```diff
@@ -0,0 +1,111 @@
+import math
```
AlexImmer, line 1: Is this model taken from somewhere or just standard? Just so people know about potential differences to a standard WRN.

AlexImmer (commented):
I recompiled the documentation to integrate the new example already. You can probably also view it locally. For detailed questions/feedback, see my comments.

wiseodd:
@AlexImmer updated. Please check.

PR #55, opened 2021-11-13
Functional laplace

metodmove:


laplace/curvature/__init__.py
```diff
@@ -13,5 +13,5 @@
     logging.info('asdfghjkl backend not available.')
 
 __all__ = ['CurvatureInterface', 'GGNInterface', 'EFInterface',
-           'BackPackInterface', 'BackPackGGN', 'BackPackEF',
+           'BackPackInterface', 'BackPackGGN', 'BackPackEF', 'BackPackGP',
```
AlexImmer, line 16: Why do the GP quantities require a new backend and cannot be added to `BackPackInterface`?

laplace/curvature/backpack.py
```diff
@@ -154,6 +154,115 @@ def kron(self, X, y, **kwargs):
         raise NotImplementedError('Unavailable through Backpack.')
 
 
+class BackPackGP(BackPackInterface):
```
AlexImmer, line 157: Overall feedback for this class:
1. make sure the docstrings are consistent (in `k_b_b` you use another style currently)
2. make the function names potentially more verbose, e.g., `gp` --> `gp_quantities`, `k_b_b` -> `kernel_batch` maybe?
3. why do we need `k_b_b` and `k_star_star`? Can they somehow be collapsed together?
4. Is the `lambdas` function redundant with something else in the code and can be combined somehow?

AlexImmer (commented):
Looks very good so far. Nice detailed tests as well. I just made two particular comments with some questions/potential considerations:

laplace/curvature/__init__.py
metodmove: I opted for having a separate class `BackPackGP`, since if we were to move gp methods (`gp`, `k_b_b` etc.) to `BackPackInterface` class, they would be available also in `BackPackGGN` and `BackpackEF` classes even though GP has nothing to do with neither GGN nor EF. The way I understand it GP is an alternative approach to Hessian approximations (GGN, EF) hence having a separate class made more sense to me.

However, this can be easily modified if you prefer to get rid of BackPackGP class. Let me know.

metodmove: Actually, you are right, `BackPackGP` class is superfluous:
- moved kernel methods to `FunctionalLaplace` class. Now they also have a simpler signature since a lot of attributes from `FunctionalLaplace` can be used instead of passing them as params
- moved `gp_quantities` to `BackPackInterface` class

laplace/curvature/backpack.py
metodmove: 1. nice catch, done
2. agreed, done
3. I thought about this, however, was not able to come up with a single function (without too many if-else branches) that could handle both cases. The difference between the two is how we handle the batch dimension: in `K_batch` we "collapse" it, while in `K_star` we want to keep the batch dimension as it is. 
4. you're right, there was duplication with `_get_full_ggn` method in `GGNInterface`. I removed `lambdas` function now and added `H_log_likelihood` in `CurvatureInterface` instead

laplace/curvature/__init__.py
AlexImmer: Sounds reasonable, I will check out the changes. I don't fully agree regarding GGN/EF not having to do anything with GP: the FullLaplace with GGN is equivalent to the full kernel GP since GGN is what leads to a GP due to linearization. In the same way, one could actually construct a GP corresponding to EF. However, this GP would only be able to predict residuals and is therefore maybe not interesting (except for estimating the marginal likelihood..).

laplace/curvature/curvature.py
```diff
@@ -89,6 +89,33 @@ def last_layer_jacobians(model, x):
 
         return Js, f.detach()
 
+    def H_log_likelihood(self, f, sigma_factor=None):
```
AlexImmer, line 92: Might be more appropriate in `GGNInterface` because only `GGN` requires this. But here would also be okay.

AlexImmer:
What is still missing for FunctionalLaplace? From what I can tell, it seems like only marginal likelihood is missing, right?

laplace/curvature/curvature.py
metodmove: if we move it to `GGNInterface`, then `BackPackInterface` needs to inherit from it. so would keep it in `CurvatureInterface` for now, let me know if you prefer otherwise

laplace/curvature/backpack.py
```diff
@@ -85,6 +85,41 @@ def gradients(self, x, y):
...
 
+    def gp_jacobians(self, X):
+        """
+        A helper function to compute jacobians (used in GP methods to reduce code duplication).
+        """
+        if self.last_layer:
+            Js, f = self.last_layer_jacobians(self.model, X)
+        else:
+            Js, f = self.jacobians(self.model, X)
+        return Js, f
```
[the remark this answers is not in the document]
metodmove: having this helper function enables me to reuse same kernel methods (`kernel_batch`, `kernel_star`, `kernel_batch_star`) in both `FunctionalLaplace` and `FunctionalLLLaplace` and hence avoid code duplication or having same if-else branch in all kernel methods. however, I agree that it's better to move it to `FunctionalLaplace` instead of having it in `BackPackInterface`, will change it

metodmove:
indeed, the only thing still missing is marginal likelihood for the classification case:
```
def _log_marginal_likelihood(self):
    self.fit(self.train_loader)
    if self.likelihood == 'classification':
        raise NotImplementedError
    elif self.likelihood == 'regression':
         return - 0.5 * (self.log_det_K + self.scatter_lml + self.M * self.n_outputs * np.log(2 * np.pi))
```

I got a bit stuck here since the algorithm from R&W 2006 does not scale linearly in nr. of classes (C) based on my understanding - I am 95% sure that there is an error (same error that you already pointed out to me). I will double check it again in the following days.

What I can do is add a support for the case where `L_diagonal=True` (L is a Hessian of log likelihood w.r.t. f, see L_{NN} in A.2.1 chapter in your [Local Linearization paper](https://arxiv.org/pdf/2008.08400.pdf)). This is anyways what we also do for a posterior in the classification case, since also for a posterior the algorithm from R&W 2006 can not be used based on my understanding. Let me know if that makes sense or if you have any other proposals here.

laplace/curvature/curvature.py
AlexImmer: That makes sense. It also doesn't hurt to have it in `CurvatureInterface` so lets leave it there.

AlexImmer:
What you say makes sense. In any case the marginal likelihood computation should be synchronized with that of the posterior predictive. So when `L_diagonal=True`, this should also be used for the marginal likelihood.

metodmove:
Asked about the issue found in R&W 2006 on CrossValidated StackExchange: https://stats.stackexchange.com/questions/555183/gaussian-processes-multi-class-laplace-approximation

metodmove:
Added marginal likelihood. In case of classification we approximate L with a diagonal matrix both in posterior as well as in marginal likelihood.

laplace/curvature/backpack.py
```diff
@@ -85,6 +85,34 @@ def gradients(self, x, y):
                         for p in self._model.parameters()], dim=1)
         return Gs, loss
 
+    def gp_quantities(self, X, y, sigma_factor):
```
AlexImmer, line 88: Could this be a part of `CurvatureInterface`, too? Then `ASDL` is supported right away if I am not mistaken.

laplace/curvature/curvature.py
AlexImmer: Ok actually I believe this should really go into`GGNInterface` and `BackPackGGN` should then be used for the FunctionalLaplace class (by default). I think it's best because this really corresponds to `GGN`, that's makes the GP come up. If we use EF, we had a different GP, and plain `BackPackInterface` is more of a interface and shouldn't be instantiated if not necessary.

AlexImmer (commented):
Looks good. I would only suggest to make the change with the backends and then rebase/merge with main and we can put this in soon. If you have time right now, we could add it in the new release.

laplace/curvature/backpack.py
metodmove: moved it to `GGNInterface`. tested `FunctionalLaplace` with `AsdlGGN` backend and it's working, see test `test_gp_equivalence_classification` in [test_functional_laplace.py](https://github.com/metodj/Laplace/blob/functional-laplace/tests/test_functional_laplace.py)

metodmove:
Addressed change requests. However, when merging latest changes from `main` branch there are quite some conflicts that will take me some time to resolve. I will resolve them over the weekend, but if you want to make a release before just go ahead 🙂

metodmove:
actually managed to resolve conflicts sooner than expected. let me know if there's anything else that I should do in case we are to include `FunctionalLaplace` into the upcoming release:

- should I run `black` formatter locally to adhere to standards mentioned in #63 ?
- is `FunctionalLaplace` too visible in the documentation (see pic below)? I guess `ParametricLaplace` and `FunctionalLaplace` have all public methods listed in the menu on the left because they are first-level children of `BaseLaplace`, is there an easy way to change that?
Screenshot 2021-12-10 at 18 22 31

laplace/lllaplace.py
```diff
@@ -100,8 +101,8 @@ def fit(self, train_loader, override=True):
                     self.model.find_last_layer(X[:1].to(self._device))
                 except (TypeError, AttributeError):
                     self.model.find_last_layer(X.to(self._device))
-            params = parameters_to_vector(self.model.last_layer.parameters()).detach()
-            self.n_params = len(params)
+            self.map_estimate = parameters_to_vector(self.model.last_layer.parameters()).detach()
```
AlexImmer, line 104: The new convention is to set the mean on initialization to `prior_mean`. Do you have a certain reason why `mean` should become `map_estimate` in LLLaplace?

laplace/laplace.py
```diff
@@ -12,15 +12,16 @@ def Laplace(model, likelihood, subset_of_weights='last_layer', hessian_structure
     likelihood : {'classification', 'regression'}
     subset_of_weights : {'last_layer', 'all'}, default='last_layer'
         subset of weights to consider for inference
-    hessian_structure : {'diag', 'kron', 'full', 'lowrank'}, default='kron'
-        structure of the Hessian approximation
+    hessian_structure : {'diag', 'kron', 'full', 'lowrank', 'GP'}, default='kron'
```
AlexImmer, line 15: Maybe name `'GP'` to `'gp'` since all others are lower-case?

laplace/baselaplace.py
```diff
@@ -63,6 +65,9 @@ def __init__(self, model, likelihood, sigma_noise=1., prior_precision=1.,
         self.n_outputs = None
         self.n_data = 0
 
+        # MAP estimate of parameters (for ParametricLaplace this corresponds to the posterior mean)
+        self.map_estimate = parameters_to_vector(self.model.parameters()).detach()
```
AlexImmer, line 69: Same comment here regarding `'mean'`.

laplace/baselaplace.py
```diff
@@ -1003,12 +1047,505 @@ def functional_variance(self, Js: torch.Tensor) -> torch.Tensor:
...
+    """
+    # key to map to correct subclass of BaseLaplace, (subset of weights, Hessian structure)
+    _key = ('all', 'GP')
+
+    def __init__(self, model, likelihood, M=None, sigma_noise=1., prior_precision=1.,
+                 prior_mean=0., temperature=1., backend=BackPackGGN, backend_kwargs=None,
+                 diagonal_kernel=False):
+        assert backend in [BackPackGGN, AsdlGGN]
+        if isinstance(backend, AsdlGGN):
+            assert likelihood == 'classification'
```
AlexImmer, line 1095: I think it would be consistent not to raise this error here but let it get raised by the backend call later as in the other laplace classes. The reason is that it is the responsibility of Asdl to know if it can implement regression (which it will be able to, soon) and hence this check might lead to unexpected behaviour once it can.

AlexImmer (commented):
Just a few nitpicky comments. Regarding the `mean` vs `map_estimate` thing: we decided to give Laplace generally  a bit more _state_: at initialization it should be just the prior distribution which allows us to apply it very nicely to continual learning. Only once `.fit()` is called, the mean will be set to the MAP. This is a very recent change, sorry that it affects your PR now a bit. I will still need to go through the details of all changes in `baselaplace.py`, this will take a bit more time either today or tomorrow.

AlexImmer:
Regarding code style, I think you followed most conventions except in some places. For now, we use `'string'` types of strings, I think you also did. Further, we try to stay below 100 linewidth but I am not sure if it's even consistent atm. We might do that automatically in #63 soon anyway so you don't need to worry so much as long as it's just linewidth and minor differences.

laplace/baselaplace.py
```diff
@@ -522,31 +581,13 @@ def __call__(self, x, pred_type='glm', link_approx='probit', n_samples=100):
...
-                kappa = 1 / torch.sqrt(1. + np.pi / 8 * f_var.diagonal(dim1=1, dim2=2))
-                return torch.softmax(kappa * f_mu, dim=-1)
-            elif link_approx == 'bridge':
-                _, K = f_mu.size(0), f_mu.size(-1)
-                f_var_diag = torch.diagonal(f_var, dim1=1, dim2=2)
-                sum_exp = torch.sum(torch.exp(-f_mu), dim=1).unsqueeze(-1)
-                alpha = 1/f_var_diag * (1 - 2/K + torch.exp(f_mu)/(K**2) * sum_exp)
-                dist = Dirichlet(alpha)
-                return torch.nan_to_num(dist.mean, nan=1.0)
+            return self._classification_predictive(f_mu, f_var, link_approx, n_samples)
```
AlexImmer, line 590: Good idea to make this a dedicated function!

laplace/baselaplace.py
```diff
@@ -1003,12 +1047,505 @@ def functional_variance(self, Js: torch.Tensor) -> torch.Tensor:
...
+
+        self.K_MM = None
+        self.Sigma_inv = None  # (K_{MM} + L_MM_inv)^{-1}
+        self.train_loader = None  # needed in functional variance and marginal log likelihood
+        self.batch_size = None
+        self.prior_factor_sod = None
+        self.mu = None  # mean in the scatter term of the log marginal likelihood
+        self.L = None
+
+    def _check_fit(self):
```
AlexImmer, line 1110: Check fit has been discarded, so can be removed.

laplace/baselaplace.py
```diff
@@ -1003,12 +1047,505 @@ def functional_variance(self, Js: torch.Tensor) -> torch.Tensor:
...
+            self.M = N
+        train_loader = self._get_SoD_data_loader(train_loader)
+        self.train_loader = train_loader
+        self.prior_factor_sod = self.M / self.n_data
+
+        self._init_K_MM()
+        self._init_Sigma_inv()
+
+        f, lambdas, mu = [], [], []
+        for i, batch in enumerate(train_loader):
```
AlexImmer, line 1201: maybe `for i, (X, y) in enumerate(train_loader)`?

laplace/baselaplace.py
```diff
@@ -1003,12 +1047,505 @@ def functional_variance(self, Js: torch.Tensor) -> torch.Tensor:
...
+            prior precision if should be changed from current `prior_precision` value
+        sigma_noise : [type], optional
+            observation noise standard deviation if should be changed
+
+        Returns
+        -------
+        log_marglik : torch.Tensor
+        """
+        # make sure we can differentiate wrt prior and sigma_noise for regression
+        self._check_fit()
```
AlexImmer, line 1348: Can remove as mentioned above.

AlexImmer (commented):
BaseLaplace also looks very good, thanks a lot. I will try it out a bit tomorrow and see if all works as expected. From the code itself, I don't see any problems.

laplace/lllaplace.py
metodmove: The reason I changed `mean` to `map_estimate` also in `BaseLaplace` and `LLLaplace` is that the attribute name `mean` is a bit misleading in the case of `FunctionalLaplace` as it does not correspond to the GP mean but rather to the `mean` of the posterior over parameters in the parametric world. However, I changed it now so that we keep `mean` in all parametric classes and `map_estimate` is used only in `FunctionalLaplace` and `FunctionalLLLaplace`

AlexImmer (commented):
I tried it out and things seem to work in the basic examples. There is some discrepancy in the `regression_example.py` script when using the marginal likelihood training procedure. Could this be numerical or is there some issue with the marginal likelihood computation potentially? If this issue is solved, we could merge this in from my side after some secondary review.

laplace/baselaplace.py
```diff
@@ -1012,8 +1054,489 @@ def sample(self, n_samples=100):
...
+        if prior_precision is not None:
+            self.prior_precision = prior_precision
+
+        # update sigma_noise (useful when iterating on marglik)
+        if sigma_noise is not None:
+            if self.likelihood != 'regression':
+                raise ValueError('Can only change sigma_noise for regression.')
+            self.sigma_noise = sigma_noise
+
+        self.fit(self.train_loader)
```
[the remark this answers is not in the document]
metodj: indeed, refitting is not needed here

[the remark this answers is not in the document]
metodj: changed the code so that refitting is not needed anymore here. note that due to this change now only isotropic priors are supported in `FunctionalLaplace`. if one uses per-parameter or per-layer priors, kernel matrices needs to be recomputed every time `prior_precision` is updated:

Screenshot 2022-12-20 at 14 37 57

added check `FucntionalLaplace` class

examples/regression_example.py
```diff
@@ -34,6 +34,7 @@ def get_model():
         optimizer.step()
 
 la = Laplace(model, 'regression', subset_of_weights='all', hessian_structure='full')
+# la = Laplace(model, 'regression', subset_of_weights='all', hessian_structure='GP')
```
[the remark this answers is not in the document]
metodj: fixed the issue with marginal likelihood, results in `regression_example.py` now look like:
- `hessian structure='gp'`:
```
sigma=0.275 |  prior precision=0.207 |  MAE: 4.526
sigma=0.277 |  prior precision=0.155 |  MAE: 4.233
```

- `hessian structure='full'`:
```
sigma=0.282 |  prior precision=0.102 |  MAE: 4.526
sigma=0.283 |  prior precision=0.153 |  MAE: 4.417
```

metodj:
Main changes since last time:

- addressed underfitting of `FunctionalLaplace` in `regression_example.py`, fixed the issue with the marginal likelihood: https://github.com/AlexImmer/Laplace/pull/55#discussion_r833521031
- added `calibration_gp_example.py` where `FunctionalLaplace` is used on a pre-trained FMNIST classifier (same experiment as in [linearised Laplace paper](https://arxiv.org/pdf/2008.08400.pdf))

metodj:
I also benchmarked the code with [BNN-predictions-repo](https://github.com/AlexImmer/BNN-predictions) in terms of speed and the code here is 2-3x slower. After investigating, I believe the main reason is that code here allows one to only use smaller batch sizes before running into memory issues when doing GP inference (`b=128` here compared to e.g. `b=512` in the BNN-predictions-repo). 

When it comes to memory, the bottleneck seem to be `torch.einsum` functions used in the kernel methods in `FunctionalLaplace`. To get around this, I also tested implementation without `torch.einsum`. However, then the memory issues started to pop up in the `backpack` backend.

Further benchmarking and faster implementation can be left for future PRs imo.

wiseodd:
Replaced by #192

Issue #16, opened 2021-06-01
Subnetwork inference


AlexImmer:
- Extend curvature backends to retrieve for subnetwork
- Add `sublaplace.py` and the corresponding classes (only `FullSubnetLaplace` possible afaik)

PR #58, opened 2021-11-30
Subnetwork Laplace

edaxberger:
This PR implements subnetwork Laplace, addressing issue #16.

As you can see from the code, this is fairly straightforward. We basically just need to ask the user to specify/pass a definition of a subnetwork. We then internally store the subnetwork as a vector of indices of the flattened/vectorized model parameters that form the subnetwork (for convenience to the user, we also allow passing the subnetwork as a binary mask/vector of the size of the parameter vector, where 1s indicate the subnetwork parameters). All we then need to do is index the Jacobians / gradients to extract the part corresponding to the subnetwork, both during inference time (i.e. for constructing the GGN/Fisher) and during prediction time (i.e. for constructing the predictive covariance matrix).

Some comments/questions:
1. I'm aware that I still need to write tests, which I'll do (I ran some tests locally to ensure that the code works generally).
2. The code also doesn't support the NN+MC predictive yet, which I'll add; it shouldn't be hard to implement, as we just need to sample the subnetwork weights and then combine them with the remaining MAP weights to form a parameter vector.
3. Should we allow the user the change the subnetwork after instantiation, or should the subnetwork remain fixed? In either case I'd need to adapt the code accordingly. For us it would be easier to not allow changing the subnetwork, as that'd require re-fitting the Hessian etc.
4. We'll have to be especially careful with the logic of the `subnetwork_mask` setter of `SubnetLaplace`, which needs to correctly check that the passed `subnetwork_mask` is in the right format (we currently support the two formats mentioned above), as the code would break otherwise. I'll double-check the logic and also add sufficient test cases for this.
5. Should we update all the internal documentation to reflect the subnetwork Laplace option? As the subnetwork Laplace implementation is integrated into the full Laplace code (and just adds lines to index the Jacobians/gradients), we'd have to change everything where we refer to the full parameter vector (incl. when we define tensor shapes, e.g. where `parameters` would become `parameters_subnet`). It's not hard to change, but might make the documentation more cluttered / harder to read, as we'd have to include conditional clauses I guess.
6. Is there are reason why `jacobians()` is static? Just wondering, as it makes the code and documentation a bit uglier/longer with the subnetwork option, as we always have to pass the `subnetwork_indices` when calling it. Compare this with `gradients()`, which is not static and can therefore simply access `self.subnetwork_indices`.
7. The code currently only supports using subnetwork Laplace with a full Hessian (and throws an error otherwise). In principle it'd also be possible to also use it with diagonal or KFAC Hessians, but it makes less sense conceptually, so I'd rather not support it. But then again, we also support options like diagonal last-layer, which are also not too sensible arguably?
8. Finally, the current implementation requires the user to specify / pass the subnetwork. We could consider also adding functionality for automatically selecting the subnetworks. Simple choices like random subnetworks or subnetworks of the largest magnitude weights would be easy to implement. But e.g. the approach we ended up using in the subnetwork inference paper (i.e. selecting weights with the largest marginal variances, where we used SWAG for variance estimation) is a bit more involved and might make the codebase too bloated. But happy to discuss this.

Thanks a lot for your efforts in looking at this (there's no rush of course)!


laplace/__init__.py
```diff
@@ -8,10 +8,13 @@
...
 from laplace.laplace import Laplace
 
 __all__ = ['Laplace',  # direct access to all Laplace classes via unified interface
            'BaseLaplace', 'ParametricLaplace',  # base-class and its (first-level) subclasses
            'FullLaplace', 'KronLaplace', 'DiagLaplace',  # all-weights
            'LLLaplace',  # base-class last-layer
-           'FullLLLaplace', 'KronLLLaplace', 'DiagLLLaplace'] # last-layer
+           'FullLLLaplace', 'KronLLLaplace', 'DiagLLLaplace',  # last-layer
+           'SubnetLaplace',  # base-class subnetwork
+           'FullSubnetLaplace']  # subnetwork
```
AlexImmer, line 20: Why are two classes required? I think for now, the `FullSubnetLaplace` could just be the default and implement exactly the algorithm as in your paper. If you could elaborate a bit on what you plan to add as subclasses, this would help me understand this choice better. From what I understand, a KronSubnet wouldn't make sense/work and diagonal neither because the diagonal is used to find the optimal subnet, right?

laplace/subnetlaplace.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @property
+    def subnetwork_mask(self):
+        return self._subnetwork_mask
+
+    @subnetwork_mask.setter
+    def subnetwork_mask(self, subnetwork_mask):
+        """Check validity of subnetwork mask and convert it to a vector of indices of the vectorized
+        model parameters that define the subnetwork to apply the Laplace approximation over.
+        """
```
AlexImmer, line 73: Maybe indicate the information present in the ValueError here in the docstring:
```
                    'Subnetwork mask needs to identify the subnetwork parameters '\
                    'from the vectorized model parameters as:\n'\
                    '1) a vector of indices of the subnetwork parameters, or\n'\
                    '2) a binary vector of size (parameters) where 1s locate the subnetwork parameters.'
```

AlexImmer:
### 1.-7. Minor comments
Looks quite nice already. I made two comments on the code directly. Here the answer/discussion to the more minor points:
1. Suggestions for tests: with all parameters retained, all should be equivalent to `FullLaplace`. With last layer retained, all should be equal to that as well.
3. The subnetwork should be static as you suggest because changing it really requires to rerun the entire `.fit()` method which is the only setup function that has a non negligible runtime.
5. It could make sense to reimplement the code separately for SubnetLaplace to make it more clear and potentially changeable in the future. Otherwise, one could add `@property` values for all class parameters that need to be masked and then override these in the `Subnet` class and mask correspondingly.
6. Yes, probably we could make the change and have `Jacobians` be an instance-method instead of static. I don't remember why it is like this right now. In fact, we have a similar change right now in the FunctionalLaplace PR: #55. If you can make the change to Jacobians in the backends on this branch, that would also work. I don't see any problem with this right now but there might be some..
7. I agree that diagonal or KFAC subnet don't make sense because KFAC would require a certain subnet structure to be sensible and the full diagonal is usually used to select the parameters of interest. I think diagonal last layer makes a little more sense but is probably also not something that should be used much..

### 8. Choice of subnetwork
One question regarding the choice of subnetwork in your paper: would it be possible to use the diagonal Laplace-GGN to choose the subnetwork? In fact, I thought that's what you did. Is there any reason that you (need to) use SWAG instead of Laplace to choose the subnetwork?

Because I thought that the subnetwork would be chosen using the diagonal GGN, I was thinking to just override the `.fit()` method with a separate run over the `data_loader` to select and set the subnetwork indices and then run the true `.fit()` computing the masked full GGN. If the method does not work with a diagonal GGN as selection criterion and we anyway also want to include different policies for selecting subnetworks, it might make sense to design a class that can be passed to the `SubnetLaplace`. One could implement it as follows
```python
class SubnetMask(object):
    def __init__(self, model, likelihood, ..):
        ...

    def __call__(self, train_loader, ...):
        # select mask based on train loader, model and other hyperparams similar to a subclass of CurvatureInterface
        return mask
```
Then, different policies can subclass this mask, and one could potentially separately implement a SWAG-Mask, random-mask, last-layer mask (for testing?). Also, the method using diagonal GGN would be easy to implement.  What do you think about this?

laplace/__init__.py
edaxberger: Yes, agreed, I removed the `SubnetLaplace` base class, making `FullSubnetLaplace` the only/default option.

edaxberger:
I now addressed most of the remaining issues discussed. Most notably, I followed your suggestion of adding a `SubnetMask` class where subclasses implement different subnetwork selection strategies. For now, random, largest magnitude and last-layer subnetworks are supported. I am considering to also add the diag-Laplace subnet selection strategy you mentioned, which should indeed also be easy to implement. I am currently finalising the tests and will push those soon as well.

edaxberger:
I also finished implementing subnet selection based on the largest marginal variances (using diagonal Laplace for variance estimation).

laplace/curvature/asdl.py
```diff
@@ -15,13 +15,12 @@
...
-    def __init__(self, model, likelihood, last_layer=False):
+    def __init__(self, model, likelihood, last_layer=False, subnetwork_indices=None):
         if likelihood != 'classification':
             raise ValueError('This backend only supports classification currently.')
-        super().__init__(model, likelihood, last_layer)
+        super().__init__(model, likelihood, last_layer, subnetwork_indices)
 
-    @staticmethod
-    def jacobians(model, x):
+    def jacobians(self, model, x):
```
AlexImmer, line 26: If jacobians become an instancemethod, they should have the same interface as `gradients` because `self` has access to model. The model does not need to be passed. This requires plenty of change all over the library, I can take care of it later.

laplace/curvature/curvature.py
```diff
@@ -59,8 +62,7 @@ def jacobians(model, x):
         """
         raise NotImplementedError
 
-    @staticmethod
-    def last_layer_jacobians(model, x):
+    def last_layer_jacobians(self, model, x):
```
AlexImmer, line 65: Same as `jacobians`

laplace/subnetlaplace.py
```diff
@@ -0,0 +1,118 @@
...
+from laplace.baselaplace import FullLaplace, DiagLaplace
+
+from laplace.curvature import BackPackGGN
+from laplace.subnetmask import LargestVarianceDiagLaplaceSubnetMask
+
+
+__all__ = ['FullSubnetLaplace']
+
+
+class FullSubnetLaplace(FullLaplace):
```
AlexImmer, line 12: Since we have just one now, I would name it `SubnetLaplace`. The most similar method is probably `LowRankLaplace` and also is implicitly _Full_.

AlexImmer: `'full'` in the key below is necessary though, as also in `LowRank`.

laplace/subnetmask.py
```diff
@@ -0,0 +1,214 @@
+import torch
+from torch.nn.utils import parameters_to_vector
+
+from laplace.feature_extractor import FeatureExtractor
+
+__all__ = ['SubnetMask', 'RandomSubnetMask', 'LargestMagnitudeSubnetMask', 'LastLayerSubnetMask', 'LargestVarianceDiagLaplaceSubnetMask']
```
AlexImmer, line 6: Formatting looks weird on github. Looks like larger indent than in the other files.

AlexImmer (commented):
Looks good, I just left some minor comments. I can help with them.

laplace/curvature/asdl.py
edaxberger: Ah yeah, good point -- I'm also happy to make that change.

laplace/subnetmask.py
edaxberger: Yes this might be because I was using tabs (= 4 spaces) instead of spaces, which the remaining code uses? I'll change to spaces.

laplace/curvature/asdl.py
edaxberger: Done.

laplace/subnetlaplace.py
edaxberger: Done

laplace/subnetmask.py
edaxberger: Done

laplace/curvature/curvature.py
edaxberger: Done

edaxberger:
I also just added an implementation of the `sample()` method for `SubnetLaplace` (e.g. to use the NN predictive), which simply only samples the weights in the subnetwork and sets all other weights to their MAP estimates (had forgotten to add that earlier). I also added tests for the predictives (which are basically identical to those from `test_baselaplace.py`).

FrederikWarburg:
Hi Erik

I'm interested in this feature! Would it be possible to specify the layer names of the subnetwork. E.g. if I want to use the Laplace approximation at layer 2 and 4 and the MAP estimate at layer 1 and 3. Could you provide an example of how I could do this?

edaxberger:
Hi @FrederikWarburg, thanks a lot for checking out our library and for your interest in the subnetwork Laplace approximation!

I agree that it would be a useful feature to be able to simply specify the names of the layers to perform the Laplace approximation over. I just implemented this functionality in two new subclasses of `SubnetMask`:
- `ParamNameSubnetMask`: allows you to specify a list of _parameter_ names (as in `model.named_parameters()`) to do inference over
- `ModuleNameSubnetMask`: the same, but with _module_ names instead (as in `model.named_modules()`); this for now only supports leaf modules (i.e. modules that have no `children()`, i.e. individual layers), as nested modules would be more complex and require a bit more thought; however, nested modules would certainly be useful to add in the future (e.g. allowing one to specify blocks of layers to perform inference over rather than having to list individual layers), as the current functionality arguably does not add much on top of `ParamNameSubnetMask`

Here's an example on how to use these:
```python
import torch
from laplace import Laplace
from laplace.subnetmask import ParamNameSubnetMask, ModuleNameSubnetMask

model = torch.nn.Sequential(nn.Linear(3, 20), nn.Linear(20, 2))
likelihood = 'classification'
dataloader = None   # define valid data loader here

# subnetwork Laplace where we specify subnetwork by parameter names
subnetwork_mask = ParamNameSubnetMask
subnetmask_kwargs = dict(parameter_names=['0.weight', '1.bias'])
lap = Laplace(model, likelihood=likelihood, subset_of_weights='subnetwork', subnetwork_mask=subnetwork_mask, hessian_structure='full', subnetmask_kwargs=subnetmask_kwargs)
lap.fit(dataloader)

# subnetwork Laplace where we specify subnetwork by module names
subnetwork_mask = ModuleNameSubnetMask
subnetmask_kwargs = dict(module_names=['0'])
lap = Laplace(model, likelihood=likelihood, subset_of_weights='subnetwork', subnetwork_mask=subnetwork_mask, hessian_structure='full', subnetmask_kwargs=subnetmask_kwargs)
lap.fit(dataloader)
```

Finally, I changed `LastLayerSubnetMask` to simply be a subclass of `ModuleNameSubnetMask` where `module_names=[last_layer_name]`.

@FrederikWarburg I hope this is useful. Please let me know in case you have any further questions or something doesn't work!

FrederikWarburg:
Awesome! Thank you for your quick reply!

I found a small error, when running on GPU. Line 59 in subnetmask.py the type is torch.cuda.ByteTensor if running on gpu. This is not part of the list and the program fails. My quick fix was just to append the cuda types to the list. Besides this, it seems to work for me!

edaxberger:
Ah, thanks a lot for pointing that out -- I indeed haven't tested the code on GPU!
I've fixed that check to be independent of the device (your fix works as well of course).

Great to hear that it works otherwise -- let me know in case there's anything else I can help with!

edaxberger:
Just made some small refactorings, and added an implementation of a subnetwork mask that selects the parameters with the largest marginal variances, where the variances are estimated using a diagonal SWAG posterior approximation (which is the approach used in the subnetwork inference paper).

The build seems to be failing now (locally everything was working fine), I'll investigate and fix this.

edaxberger:
Ah ok, this was due to the recent changes introduced by https://github.com/AlexImmer/Laplace/pull/72. I merged the latest `main` into `subnetlaplace` and fixed the issue.

edaxberger:
One question might be how to structure the new files. This PR adds three new files (`subnetlaplace.py`, `subnetmask.py` and `swag.py`), which are currently simply in `laplace/`. Perhaps it makes sense to put (a subset of) these into a dedicated sub-directory (e.g. `laplace/subnetlaplace/` or `laplace/subnetwork/`)?

runame:
> One question might be how to structure the new files. This PR adds three new files (`subnetlaplace.py`, `subnetmask.py` and `swag.py`), which are currently simply in `laplace/`. Perhaps it makes sense to put (a subset of) these into a dedicated sub-directory (e.g. `laplace/subnetlaplace/` or `laplace/subnetwork/`)?

Other options would be:
1. Leave `subnetlaplace.py` and `subnetmask.py`, move content of `swag.py` into `utils.py` or create new `utils` directory with `utils.py` and `swag.py`.
2. Also create a `utils` directory, but move `subnetmask.py`, `matrix.py`, and `feature_extractor.py` there as well.

edaxberger:
> 2. Also create a `laplace/utils` directory, but move `subnetmask.py`, `matrix.py`, and `feature_extractor.py` there as well.

I like this option, as it would make `laplace` a bit cleaner, but no strong preference.

edaxberger:
On another note, I was just going to update the README to include subnetwork Laplace, and realized that low-rank Laplace is also not yet mentioned. Also, we might want to update the paper citation to mention NeurIPS. Happy to quickly make these changes.

runame:
> I like this option, as it would make `laplace` a bit cleaner, but no strong preference.

I think this is also my preferred option. @AlexImmer?

runame:
I went through the code and it looks very good already! I have one main point to discuss:

Currently, the `SubnetMask` child class is passed to `Laplace` for initialization, along with `subnetmask_kwargs`; this follows the same flow used for the `backend` class. However, this results in some inconveniences, most notably that the subnetwork is only selected when calling `fit()` because only at this point the `train_loader` is available. If the selection was happening outside the `fit()` method, it would also be possible to use subnet Laplace for continual learning with the standard interface, i.e. the posterior is initialized to the prior during initialization and the Hessian is accumulated with the `override=False` keyword to `fit()` (which is currently still missing and needs to be added in any case).
So an alternative would be to keep the whole subnet selection process separate from Laplace, i.e. instantiate the `SubnetMask` child class outside of the Laplace class with all necessary keywords and the train loader, and also call the `select` method. I think then we only need to pass the subnet indices to `Laplace`, which also makes the code in `SubnetLaplace` a bit cleaner. Since the selection method and the keywords have to be manually selected by the user outside of the Laplace class anyway, I think just doing everything outside does not add much overhead for the user.

Minor notes:
- Structure of `laplace` directory, as mentioned above.
- Cost of Jacobians, but there is no easy and general fix I think.

edaxberger:
Thanks for looking over the code and your suggestion, Runa!

You're raising a good point. I was also considering outsourcing the subnetwork mask instantiation/selection to the user, but in the end decided for the current design as it makes the user interface a little bit simpler IMO. But I agree that it might be a bit cleaner if we keep the subnetwork selection separate (and just pass either the `SubnetMask` instance after subnet selection or directly the subnetwork indices to `Laplace()`). In particular, I haven't considered the continual learning setting, which is a good reason to use this alternative design.

Happy to make that change. @AlexImmer what do you think?

> Cost of Jacobians, but there is no easy and general fix I think.
By this you mean the issue that we compute the full Jacobians (over all parameters) before selecting the relevant entries corresponding to the subnetwork parameters, rather than only computing the required Jacobian entries in the first place (which should be much more efficient in theory)? I agree -- this is something I have tried to figure out for some time but couldn't find a solution for. AFAIK the PyTorch autograd engine can only compute gradients w.r.t. full parameter tensors but not sub-tensors, unfortunately (see e.g. https://discuss.pytorch.org/t/how-to-turn-off-requires-grad-for-individual-params/103303/8). Would be great to find a solution to this at some point, as it could significantly speed things up for large models.

edaxberger:
I've now also updated the README to include both subnetwork Laplace and low-rank Laplace, and updated the paper citation to the NeurIPS version -- feel free to take a look.

@AlexImmer What were the papers that you based the low-rank Laplace implementation on? For now I've just mentioned [Sharma et al. 2021](https://arxiv.org/abs/2102.12567) in the reference list in the README, but there's e.g. also [Maddox et al. 2020](https://arxiv.org/abs/2003.02139) that might be relevant. We might also want to consider citing the relevant papers in the doc string of `LowRankLaplace` (as we e.g. also do for the marginal likelihood training script and `SubnetLaplace`).

AlexImmer:
- Re structuring masks, utils, feature_extractor, matrix, etc. into `utils/` makes sense but subnet could also have its own directory `subnet/` with corresponding utilities inside. I would be fine with both.
- Re passing masks only: I think this is a good idea if it works as expected. Would it be a problem that the backend requires fitting the mask before initializing laplace? This is necessary because the backend requires the subnet-indices, right?
- Re references: I just implemented one way of doing subnet-laplace using the eigendecomposition. I think the Sharma reference does almost the same so let's use that one. Maddox et al don't propose a low-rank LA. In the future, we should probably add references to the docstrings, I will make an issue for that. It's a separate thing.
- Re jacobians: I also thought it would be doable more efficiently but couldn't find any way for the general case when they are fully indexed.
- @FrederikWarburg thanks for spotting the device error!

edaxberger:
Thanks for your comments, Alex!
I'll address the remaining issues next week, then we can hopefully merge this in and ship the new release by the end of the year (or whenever you guys are available)?

> * Re structuring masks, utils, feature_extractor, matrix, etc. into `utils/` makes sense but subnet could also have its own directory `subnet/` with corresponding utilities inside. I would be fine with both.

Yes that's also an option. I'll try both options to see which one feels better.

> * Re passing masks only: I think this is a good idea if it works as expected. Would it be a problem that the backend requires fitting the mask before initializing laplace? This is necessary because the backend requires the subnet-indices, right?

Yes exactly, we'd need to fit the mask before passing it to `Laplace()`, as the backend will need the subnet indices. I don't think this is an issue, as it's the same way we do it now -- we'd just outsource the mask fitting to the user and not do it internally before fitting Laplace. But I might also have misunderstood your concern. In the end using subnet Laplace would look something like this:
```python
subnetwork_mask = SubnetMask(**subnetmask_kwargs) # instantiate subnet mask
subnetwork_mask.select(train_loader) # select subnetwork
lap = Laplace(model, likelihood, subset_of_weights='subnetwork', subnetwork_mask=subnetwork_mask)
```
Then internally we just need to pass `subnetwork_mask.indices` to the backend (probably simply in the `Laplace()` constructor) and we're good to go.

> * Re references: I just implemented one way of doing subnet-laplace using the eigendecomposition. I think the Sharma reference does almost the same so let's use that one. Maddox et al don't propose a low-rank LA. In the future, we should probably add references to the docstrings, I will make an issue for that. It's a separate thing.

Sounds good!

runame:
End of year release sounds good!

- Re `laplace` directory structure: my preferred option is a `utils` directory, simply because it makes everything cleaner and we already have files which would fit well in there, e.g. `matrix.py` and `feature_extractor.py`.
- Re keeping the `SubnetMask` logic external: is it also possible to only pass `subnetwork_mask.indices` to `Laplace`? I think nothing else is needed (the `n_params_subnet` attribute can be derived from it)  and this enables users to choose indices even without using the `SubnetMask` classes or writing their custom subclass.

edaxberger:
I have now implemented the remaining things we had discussed, namely:
- I moved all utility files (`feature_extractor.py`, `matrix.py`, `subnetmask.py`, `swag.py`, `utils.py`) into a new `laplace/utils` folder
- I changed the logic of `SubnetLaplace` to take `subnetwork_indices` instead of a subclass of `SubnetMask`; this includes careful checks for the validity of the passed `subnetwork_indices`
- I updated the tests accordingly (i.e. by adapting existing tests to the new structure and adding new tests for custom `subnetwork_indices` not derived from any `SubnetMask` subclass)

Let me know what you think and if there is anything else that I should take care of!

edaxberger:
I also just added an example for using `SubnetLaplace` (with different ways to choose the `subnetwork_indices`) to the README.

laplace/subnetlaplace.py
```diff
@@ -0,0 +1,129 @@
+import torch
+from torch.distributions import MultivariateNormal
+
+from laplace.baselaplace import FullLaplace
+from laplace.curvature import BackPackGGN
+
+
+__all__ = ['SubnetLaplace']
```
AlexImmer, line 8: I think It's good practice to have this in all files. Of course that's out of scope for this PR but maybe we should make an issue for this? Do you know what exactly this helps with?

laplace/subnetlaplace.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        if subnetwork_indices is None:
+            raise ValueError('Subnetwork indices cannot be None.')
+        elif not (isinstance(subnetwork_indices, torch.LongTensor) and\
+            subnetwork_indices.numel() > 0 and len(subnetwork_indices.shape) == 1):
+            raise ValueError('Subnetwork indices must be non-empty, 1-dimensional torch.LongTensor.')
+        elif not (len(subnetwork_indices[subnetwork_indices < 0]) == 0 and\
+            len(subnetwork_indices[subnetwork_indices >= self.n_params]) == 0):
+            raise ValueError(f'Subnetwork indices must lie between 0 and n_params={self.n_params}.')
+        elif not (subnetwork_indices.sort()[0].equal(torch.unique(subnetwork_indices, sorted=True))):
```
AlexImmer, line 99: `len(subnetwork_indices.unique()) == len(subnetwork_indices)` is probably a bit better to read and would do the same?

laplace/utils/subnetmask.py
```diff
@@ -0,0 +1,347 @@
+from copy import deepcopy
+
+import torch
+from torch.nn import CrossEntropyLoss, MSELoss
+from torch.nn.utils import parameters_to_vector
+
+from laplace.utils.feature_extractor import FeatureExtractor
+from laplace.utils.swag import fit_diagonal_swag
+
+__all__ = ['SubnetMask', 'RandomSubnetMask', 'LargestMagnitudeSubnetMask', 'LargestVarianceDiagLaplaceSubnetMask', 'LargestVarianceSWAGSubnetMask', 'ParamNameSubnetMask', 'ModuleNameSubnetMask', 'LastLayerSubnetMask']
```
AlexImmer, line 10: Indent missing

AlexImmer, line 10: Indent missing

laplace/utils/swag.py
```diff
@@ -0,0 +1,81 @@
...
+
+import torch
+from torch.nn.utils import parameters_to_vector
+
+
+def param_vector(model):
+    return parameters_to_vector(model.parameters()).detach()
+
+
+def fit_diagonal_swag(model, train_loader, criterion, n_snapshots_total=40, snapshot_freq=1, lr=0.01, momentum=0.9, weight_decay=3e-4, min_var=1e-30):
```
AlexImmer, line 11: Just out of interest: the mean is just discarded and not useful for the final LA or could it help to get a better mean following the idea of standard SWA? The function name indicates it does the full `swag` (mean + var) currently. If you just need the variance, maybe that can be indicated in the function name like `fit_diagonal_swag_var`. But it's not necessary, just a suggestion.

AlexImmer (approved):
Great work, looks good to me now. I just left a few comments/ questions. Other than that 
- It would be great if you could make sure that the linewidth does not exceed (90? 100?). Maybe we didn't agree on the linewidth, yet. I'd suggest to go with 100 for now. 
- Did you recompile the documentation as indicated in the readme? Otherwise I can do that if you want.

laplace/subnetlaplace.py
runame: I think it determines what is imported when you call
```python
from laplace.subnetlaplace import *
```
and maybe there are other effects too.

AlexImmer:
Probably also good for autocomplete?

laplace/utils/subnetmask.py
```diff
@@ -0,0 +1,347 @@
...
+        subnet_mask_indices : torch.LongTensor
+            a vector of indices of the vectorized model parameters
+            (i.e. `torch.nn.utils.parameters_to_vector(model.parameters())`)
+            that define the subnetwork
+        """
+        if not isinstance(subnet_mask, torch.Tensor):
+            raise ValueError('Subnetwork mask needs to be torch.Tensor!')
+        elif subnet_mask.dtype not in [torch.int64, torch.int32, torch.int16, torch.int8, torch.uint8, torch.bool] or\
+                len(subnet_mask.shape) != 1:
+                raise ValueError('Subnetwork mask needs to be 1-dimensional integral or boolean tensor!')
```
runame, line 65: Indent too big.

laplace/lllaplace.py
```diff
@@ -117,7 +117,7 @@ def fit(self, train_loader, override=True):
         self.mean = parameters_to_vector(self.model.last_layer.parameters()).detach()
 
     def _glm_predictive_distribution(self, X):
```
runame, line 119: This method can be removed with a minor refactor in some places, but I'll just do that in a later PR.

laplace/subnetlaplace.py
```diff
@@ -0,0 +1,129 @@
...
+    backend_kwargs : dict, default=None
+        arguments passed to the backend on initialization, for example to
+        set the number of MC samples for stochastic approximations.
+    """
+    # key to map to correct subclass of BaseLaplace, (subset of weights, Hessian structure)
+    _key = ('subnetwork', 'full')
+
+    def __init__(self, model, likelihood, subnetwork_indices, sigma_noise=1., prior_precision=1.,
+                 prior_mean=0., temperature=1., backend=BackPackGGN, backend_kwargs=None):
+        self.H = None
```
runame, line 75: It looks like `_init_H()` will only be called when `fit()` is called with `override=True`, so it would fail for `override=False`, or am I missing something? Maybe it's worth adding a test for that (and maybe also the test for initialization with a large model, but this should work here, since you correctly set `self.H = None` before inheriting anything).

runame (commented):
Looks great! See my last comment for my only concern.

Minor notes:
- Regarding the line length, I also think 100 is fine for now. I think we have always used brackets `([{` for line breaks, so it would be nice to be consistent with that.
- We could consider also moving `marglik_training.py` to the new `utils` directory (or just leave it where it is or in the future add something like a `methods` directory).

laplace/subnetlaplace.py
```diff
@@ -0,0 +1,129 @@
...
+    def _init_H(self):
+        self.H = torch.zeros(self.n_params_subnet, self.n_params_subnet, device=self._device)
+
+    def _check_subnetwork_indices(self, subnetwork_indices):
+        """Check that subnetwork indices are valid indices of the vectorized model parameters
+           (i.e. `torch.nn.utils.parameters_to_vector(model.parameters())`).
+        """
+        if subnetwork_indices is None:
+            raise ValueError('Subnetwork indices cannot be None.')
+        elif not (isinstance(subnetwork_indices, torch.LongTensor) and\
```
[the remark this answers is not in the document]
AlexImmer: That's fairly standard in torch somehow. For example, cross-entropy loss also would expect a long tensor afaik.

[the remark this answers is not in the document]
runame: When I wrote the comment I actually asked myself the exact same thing, i.e. why is this also the case for the PyTorch cross-entropy loss, and even did a quick google search but didn't find anything.

AlexImmer:
> * Regarding the line length, I also think 100 is fine for now. I think we have always used brackets `([{` for line breaks, so it would be nice to be consistent with that.
> * We could consider also moving `marglik_training.py` to the new `utils` directory (or just leave it where it is or in the future add something like a `methods` directory).

- Agree with the first comment. I think in the PR are plenty of linebreaks backslash.
- I wouldn't put it into `utils`. I think `utils` should only contain utilities that are required for the package and not utilities for users. The latter we can call `methods` later as discussed.

laplace/baselaplace.py
```diff
@@ -4,8 +4,8 @@
 from torch.nn.utils import parameters_to_vector, vector_to_parameters
 from torch.distributions import MultivariateNormal, Dirichlet, Normal
 
-from laplace.utils import parameters_per_layer, invsqrt_precision, get_nll, validate
-from laplace.matrix import Kron
+from laplace.utils.utils import parameters_per_layer, invsqrt_precision, get_nll, validate
+from laplace.utils.matrix import Kron
```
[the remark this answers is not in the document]
edaxberger: Agreed, done.

laplace/subnetlaplace.py
edaxberger: Yes good point -- I added `__all__` for the all the files in `utils/` for now, and will create an issue for any remaining files.

laplace/subnetlaplace.py
edaxberger: Yes of course, good catch, changed it!

laplace/utils/subnetmask.py
edaxberger: Fixed

laplace/utils/swag.py
edaxberger: Good point! iirc I have tried using the SWA mean with Laplace, which indeed improved performance. But I haven't done extensive experiments with it as I thought it's not as principled (although perhaps one could view the SWA mean as some kind of approximation to the MAP?), but it would be interesting to revisit the idea (and perhaps even add this as a feature if it works well). The only downside I see is that it requires us to update the BatchNorm parameters of the model, which is one of the main reasons why SWA(G) is so slow.

I've changed the name to `fit_diagonal_swag_var` for clarity, thanks!

laplace/subnetlaplace.py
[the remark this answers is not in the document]
edaxberger: Yeah I did it for direct compatibility with PyTorch (which requires indices for tensors to be of type `torch.LongTensor`). Also not sure why PyTorch is using this design in many places, but I could indeed assume that it's to prophylactically avoid pesky overflow issues when scaling things up in the future. In our specific case, I think using int64 does make sense, given that modern models can indeed have more parameters than the int32 limit (although still rare these days). Also, `torch.tensor([1, 2, 3])` seems to return a `torch.LongTensor` by default, so most people would be dealing with int64 tensors anyways if they don't explicitly cast to a lower precision type. But I'm also fine with allowing any integral type and internally converting to int64 if you prefer that (would indeed be more flexible).

laplace/lllaplace.py
edaxberger: Ah ok, cool!

laplace/subnetlaplace.py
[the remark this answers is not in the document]
runame: I'm fine with leaving it how it is, since the default integer tensor is `torch.LongTensor` most users probably won't run into issues anyway.

laplace/utils/subnetmask.py
edaxberger: Fixed, thanks!

laplace/subnetlaplace.py
edaxberger: Good catch! I believe nothing speaks against calling `_init_H()` at the end of the constructor here, since we now always know `self.n_params_subnet` from the very beginning when passing `subnetwork_indices`, so I'll just add that in to make `fit()` work with `override=True` from the very beginning.

I'll also add the test with the large model!

edaxberger:
Thanks a lot for your comments, Alex and Runa, which I have now addressed! I also recompiled the documentation. Let me know if there is anything else that needs to be resolved.

edaxberger:
Just realised that for some reason Travis has not been running the tests here anymore for the latest commits (I ran tests locally, where everything worked). Any idea why it's not doing this check anymore?

runame:
> Just realised that for some reason Travis has not been running the tests here anymore for the latest commits (I ran tests locally, where everything worked). Any idea why it's not doing this check anymore?

We have limited compute available on Travis, so we only run the tests when merging PRs from now on. If you ran the tests locally it should be fine.

runame (approved):

edaxberger:
> We have limited compute available on Travis, so we only run the tests when merging PRs from now on. If you ran the tests locally it should be fine.

Ah I see, makes sense!

I just merged `main` into `subnetlaplace` and re-ran the tests locally, and everything still works. So feel free to merge this into `main`!

PR #60, opened 2021-12-05
Minor improvements

runame:
- Handles the case when the inputs `X` do not support slicing or the slices are not PyTorch tensors, i.e. have no `.to(device)` function.
- Does not require `X` to have `__len__()` implemented.
- Creates a deep copy of the model for last-layer Laplace. I think this is better because the `FeatureExtractor` wrapper uses hooks, which means that the model is modified. This might lead to unexpected behavior for the user.

There are still a lot of requirements on the given `train_loader` and its batches which have to be handled by the user by creating appropriate wrappers for their data loaders. Maybe we can lessen the requirements more in the future.


AlexImmer:
What kind of unexpected behaviour can come up when no copying the FeatureExtractor? Because it also doubles the size of the model right? Normally that shouldn't be a problem though, I guess.

AlexImmer (approved):

runame:
I'm actually not completely sure anymore what the unexpected behavior was that I noticed, but I suspect it might have been something like [this issue](https://github.com/pytorch/pytorch/issues/1148). Whatever it was, it raised my awareness for the fact that the underlying model is changed, which I think can be considered bad design in general, i.e. we don’t know what issues a user might encounter because of it. In most cases the additional memory shouldn’t be a problem, but I’m also fine with leaving it without the copy if you think the trade-off is not worth it.

PR #62, opened 2021-12-07
Support for refitting LA with option to override or update

AlexImmer:
Changes:
- make `mean` a quantity that falls back to prior when not yet fit and is set on `.fit()` so that `.log_prob()` is correct
- add `log_prob(value)` to compute log density under LA for regularization. This falls back to prior before fitting the LA
- add `.fit(train_loader, override=True)` where `override` means to reset Hessian approximation while `override == False` leads to adding up curvature, e.g., for continual learning

### DIFF
--- a/laplace/baselaplace.py
+++ b/laplace/baselaplace.py
@@ -181,7 +181,7 @@ class BaseLaplace:
     def optimize_prior_precision_base(self, pred_type, method='marglik', n_steps=100, lr=1e-1,
                                       init_prior_prec=1., val_loader=None, loss=get_nll,
                                       log_prior_prec_min=-4, log_prior_prec_max=4, grid_size=100,
-                                      link_approx='probit', n_samples=100, verbose=False, 
+                                      link_approx='probit', n_samples=100, verbose=False,
                                       cv_loss_with_var=False):
         """Optimize the prior precision post-hoc using the `method`
         specified by the user.
@@ -334,11 +334,11 @@ class ParametricLaplace(BaseLaplace):
         try:
             self._init_H()
         except AttributeError:  # necessary information not yet available
-            pass  
+            pass
         self.loss = 0
         self.n_data = 0
         # posterior mean/mode
-        self.mean = self.prior_mean 
+        self.mean = self.prior_mean
 
     def _init_H(self):
         raise NotImplementedError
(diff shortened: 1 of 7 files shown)


laplace/baselaplace.py
```diff
@@ -336,34 +331,37 @@ def __init__(self, model, likelihood, sigma_noise=1., prior_precision=1.,
...
                          prior_mean, temperature, backend, backend_kwargs)
-
-        self.H = None
-
+        try:
+            self._init_H()
+        except AttributeError:  # necessary information not yet available
+            pass
+        self.loss = 0
+        self.n_data = 0
```
runame, line 339: `loss` and `n_data` attributes are already initialized in the `BaseLaplace` class.

laplace/baselaplace.py
```diff
@@ -763,13 +806,22 @@ def posterior_precision(self):
...
         """
-        self._check_fit()
+        if type(self.H) is Kron:  # Fall back to prior
+            return self.prior_precision_diag
         return self.H * self._H_factor + self.prior_precision
 
     @property
     def log_det_posterior_precision(self):
+        if type(self.H) is Kron:  # Fall back to diag prior
+            return self.posterior_precision.log().sum()
```
runame, line 816: This adds one redundant `if` statement in `posterior_precision`. Could be directly replaced with `prior_precision_diag.log().sum()`.

laplace/lllaplace.py
```diff
@@ -65,32 +65,32 @@ def __init__(self, model, likelihood, sigma_noise=1., prior_precision=1.,
...
+            self.mean = prior_mean
             self.n_params = None
             self.n_layers = None
             # ignore checks of prior mean setter temporarily, check on .fit()
             self._prior_precision = prior_precision
             self._prior_mean = prior_mean
         else:
-            self.mean = parameters_to_vector(self.model.last_layer.parameters()).detach()
-            self.n_params = len(self.mean)
+            self.mean = prior_mean
```
runame, line 75: I think this line should be `self.mean = self.prior_mean` after line 79 so the prior mean setter checks also apply.

runame (commented):
Looks great, this really simplifies continual learning! I wrote a few minor comments.

Optional, probably not needed:
- Test repeated fit for last-layer Laplace
- Test initialization of mean and precision to prior

laplace/baselaplace.py
AlexImmer: That would still be necessary unless we also change it in `square_norm`. But I will do this because then it's clear that it is falling back to the prior.

runame: I see, thanks.

PR #63, opened 2021-12-07
[WIP] Proposal for black code style

aleximmer:
We don't really have any style guide and this would be the easiest as it's auto-enforced by `black laplace` after installing it. 

To discuss maybe:
- default 100 linewidth?
- strings with `'` instead of `"` (this wasn't even consistent before in our code)
- should tests and examples be formatted like this?

What do you think?

### DIFF
--- a/laplace/utils.py
+++ b/laplace/utils.py
@@ -12,16 +12,15 @@ def get_nll(out_dist, targets):
 
 
 @torch.no_grad()
-def validate(laplace, val_loader, pred_type='glm', link_approx='probit', n_samples=100):
+def validate(laplace, val_loader, pred_type="glm", link_approx="probit", n_samples=100):
     laplace.model.eval()
     output_means, output_vars = list(), list()
     targets = list()
     for X, y in val_loader:
         X, y = X.to(laplace._device), y.to(laplace._device)
         out = laplace(
-            X, pred_type=pred_type,
-            link_approx=link_approx,
-            n_samples=n_samples)
+            X, pred_type=pred_type, link_approx=link_approx, n_samples=n_samples
+        )
 
         if type(out) == tuple:
             output_means.append(out[0])
@@ -33,8 +32,10 @@ def validate(laplace, val_loader, pred_type='glm', link_approx='probit', n_sampl
 
     if len(output_vars) == 0:
         return torch.cat(output_means, dim=0), torch.cat(targets, dim=0)
-    return ((torch.cat(output_means, dim=0), torch.cat(output_vars, dim=0)),
-            torch.cat(targets, dim=0))
+    return (
+        (torch.cat(output_means, dim=0), torch.cat(output_vars, dim=0)),
+        torch.cat(targets, dim=0),
+    )
 
 
 def parameters_per_layer(model):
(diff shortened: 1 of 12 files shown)


laplace/utils.py
```diff
@@ -103,9 +106,9 @@ def kron(t1, t2):
     tiled_t2 = t2.repeat(t1_height, t1_width)
     expanded_t1 = (
         t1.unsqueeze(2)
-          .unsqueeze(3)
-          .repeat(1, t2_height, t2_width, 1)
-          .view(out_height, out_width)
+        .unsqueeze(3)
```
wiseodd, line 109: The old style for this looks better IMO

laplace/utils.py
```diff
@@ -183,6 +186,6 @@ def block_diag(blocks):
     p_cur = 0
     for block in blocks:
         p_block = block.shape[0]
-        M[p_cur:p_cur+p_block, p_cur:p_cur+p_block] = block
+        M[p_cur : p_cur + p_block, p_cur : p_cur + p_block] = block
```
wiseodd, line 189: I think the additional spaces here make it harder to parse, but maybe I'm just unfamiliar to this style.

wiseodd:
* Linewidth 100 sounds about right to me (I usually use 120)
* About string with `'` or `"`, I'm used to the former. But I don't have any preference---PEP8 says that you can pick whichever, as long as we're consistent
* Formatting test: I think it's useful since it's part of the library's codebase
* Formatting examples: I think it's unnecessary and just adding extra lines

edaxberger:
I was about to suggest this as well -- I have been using black in the past and liked it (although I agree that in some cases, it might produce formatting that a human wouldn't come up with).

> default 100 linewidth?
Sure, fine with that.

> strings with ' instead of " (this wasn't even consistent before in our code)
I personally have a slight preference for ' (as it's a bit more minimalistic), but no strong opinion here.

> should tests and examples be formatted like this?
Agreed with Agustinus that it's fine for tests, but might make examples look weird / more verbose. But we can just try for our examples and see if we like it.

edaxberger:
Perhaps another thing to discuss: how should we make sure this style is used?

I think there's several options:
- just trusting contributors on running their code through black before committing code
- requiring contributors to use a pre-commit hook to automatically run black before every commit (this is what I had done in my previous projects)
- one can also use GitHub Actions to automatically apply the formatting after every push or PR (e.g. using something like [this](https://github.com/rickstaa/action-black/)), but I haven't used Actions yet and am therefore not familiar with it

runame:
> default 100 linewidth?

I think that's fine. If not, I'd rather have something like 90 than something > 100.

> strings with ' instead of " (this wasn't even consistent before in our code)

I'm in favor of `'`.

> should tests and examples be formatted like this?

I don't see a reason why not. In the case of the examples, we can check if it hurts readability.

In general, I also think that this code style sometimes looks a bit weird, e.g. in the cases Agustinus pointed out. In any case, I think it makes sense to not include this in the NeurIPS release, so we have a bit more time to decide on the details; it doesn't change the functionality anway. Also, I think using GitHub Actions sounds like a good way to enforce whatever code style we decide on.

aleximmer:
Closing since we decided to use ruff instead #152 and have it added in #144

PR #64, opened 2021-12-09
Online marglik as a method

AlexImmer:


laplace/marglik.py
```diff
@@ -0,0 +1,248 @@
...
+            else:
+                crit_factor = temperature
+            prior_prec = torch.exp(log_prior_prec).detach()
+            theta = parameters_to_vector(model.parameters())
+            delta = expand_prior_precision(prior_prec, model)
+            f = model(X)
+            loss = criterion(f, y) + (0.5 * (delta * theta) @ theta) / N / crit_factor
+            loss.backward()
+            optimizer.step()
+            epoch_loss += loss.cpu().item() / len(train_loader)
```
runame, line 189: The loss is not exact since the size of the last batch is typically smaller than the batch size, i.e. the correct thing would be to multiply the loss of every batch by `len(X)` and then devide by `N` at the end.

laplace/marglik.py
```diff
@@ -0,0 +1,248 @@
...
+def marglik_training(
+    model,
+    train_loader,
+    likelihood='classification',
+    optimizer_cls=Adam,
+    optimizer_kwargs=None,
+    scheduler_cls=None,
+    scheduler_kwargs=None,
+    n_epochs=300,
+    laplace=KronLaplace,
```
runame, line 23: Maybe it makes sense to use the `Laplace` function interface, since we use it everywhere else as the UI. So here you would pass the `laplace_kwargs` or `subset_of_weights`, `hessian_structure`, and `laplace_kwargs`.

runame (commented):
Looks good! I think my first comment below regarding the `Laplace` interface is the only thing to discuss.

One more very minor question: any reason why you did not call the file `marglik_training.py`?

runame:
Not sure what we decided before, but we also considered having a designated directory for methods, right? We can still add that later when we have more methods though.

laplace/marglik.py
AlexImmer: Good catch, thanks. I fixed it.

laplace/marglik_training.py
```diff
@@ -0,0 +1,259 @@
...
+            else:
+                crit_factor = temperature
+            prior_prec = torch.exp(log_prior_prec).detach()
+            theta = parameters_to_vector(model.parameters())
+            delta = expand_prior_precision(prior_prec, model)
+            f = model(X)
+            loss = criterion(f, y) + (0.5 * (delta * theta) @ theta) / N / crit_factor
+            loss.backward()
+            optimizer.step()
+            epoch_loss += loss.cpu().item() / len(train_loader)
```
runame, line 194: If I'm not missing something you now devide the loss by `N` in line 202, but you still have to multiply by `len(X)` here instead of divide by `len(train_loader)`.

laplace/marglik.py
AlexImmer: Done as suggested, thanks.

laplace/marglik_training.py
AlexImmer: Oh yes, thanks. I mixed it up with `epoch_perf`...

runame (approved):

PR #65, opened 2021-12-10
Low rank LA

AlexImmer:
Adds low-rank LA and true diagonal Hessian backend using ASDL.


laplace/baselaplace.py
```diff
@@ -664,6 +663,100 @@ def posterior_precision(self):
...
+        if not override:
+            # LowRankLA cannot be updated since eigenvalue representation not additive
+            raise ValueError('LowRank LA does not support updating.')
+
+        self.model.eval()
+        self.mean = parameters_to_vector(self.model.parameters()).detach()
+
+        X, _ = next(iter(train_loader))
+        with torch.no_grad():
+            self.n_outputs = self.model(X[:1].to(self._device)).shape[-1]
```
runame: This should be updated to be consistent with the try-except statement we use everywhere else in `fit()`.

runame (commented):
Besides my one minor comment everything looks good now.

PR #71, opened 2021-12-13
Refactor to keep initialization of H for all-weights and last-layer separate

runame:
Addresses bug resulting from #62, first reported [here](https://github.com/runame/laplace-experiments/pull/17) (note: links to PR discussion in private repository). For the last-layer LA flavors the Hessian approximation `H` was first initialized for all-weights, which leads to out-of-memory errors for larger models.

Changes:
- Add `ALLaplace`, a base-class for all all-weights flavors, i.e. what `LLLaplace` is to the last-layer flavors. Through this change we can keep the initialization of `H` for the all-weights flavors completely separate from the initialization for the last-layer flavors.
- Refactor the all-weights flavors to have base-classes for `hessian_structure` flavors `'full'`, `'kron'`, and `'diag'`, which are used by their all-weights and last-layer variations of the corresponding Hessian structure.
- Remove redundant code + minor fixes.
- Add tests of initialization with a Wide ResNet 50-2 model, to check if it fails for `FullLaplace` and works with all last-layer flavors.

I think this refactor requires careful review, to avoid introducing some different unintended behavior.

### DIFF
--- a/laplace/__init__.py
+++ b/laplace/__init__.py
@@ -7,13 +7,17 @@
 REGRESSION = 'regression'
 CLASSIFICATION = 'classification'
 
-from laplace.baselaplace import BaseLaplace, ParametricLaplace, FullLaplace, KronLaplace, DiagLaplace, LowRankLaplace
+from laplace.baselaplace import (
+    BaseLaplace, ParametricLaplace, FullLaplaceBase, KronLaplaceBase, DiagLaplaceBase,
+    ALLaplace, FullLaplace, KronLaplace, DiagLaplace, LowRankLaplace)
 from laplace.lllaplace import LLLaplace, FullLLLaplace, KronLLLaplace, DiagLLLaplace
 from laplace.laplace import Laplace
 from laplace.marglik_training import marglik_training
 
 __all__ = ['Laplace',  # direct access to all Laplace classes via unified interface
            'BaseLaplace', 'ParametricLaplace',  # base-class and its (first-level) subclasses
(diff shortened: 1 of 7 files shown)


runame:
@edaxberger It would be great if you could test if this fixes the issue in the WILDS experiment.

edaxberger:
Thanks Runa, looks great to me! I tested it with WILDS and it works properly now.

One minor comment with regards to naming:

For clarity and consistency with the LL classes (e.g. `FullLLLaplace`), should we also include the `AL` in the all-layer class names (e.g. `FullALLaplace` instead of `FullLaplace`)?

As it would make the class names a bit harder to read (IMO the LL ones are already a bit hard to parse), we could also consider writing things out, i.e. replacing `AL` by `AllLayer` and `LL` by `LastLayer`, e.g. `FullLastLayerLaplace` instead of `FullLLLaplace`.

I would be in favour of that change as it would improve clarity in my view, but I also understand that it would make the class names quite long (which I guess was your motivation for using `LL` rather than `LastLayer`?).

runame:
I actually just wrote a different fix to avoid all the new classes... My idea before was to cleanly separate the all-weights and last-layer classes, such that bugs like the one with `_H_init()` are less likely to occur in the future. The issue is that this creates many more classes, so after some discussions with Alex we decided that it's probably not worth the additional complexity. I will create a different PR with my alternative fix. Let us know what you think!

edaxberger:
Ah great, even better if you found a clean solution that doesn't require all the new classes -- looking forward to taking a look!

Issue #69, opened 2021-12-13
Add tests with larger model architectures


edaxberger:
We should add some tests to catch potential issues with larger models. When computing Hessians it's very easy to run out of memory on consumer hardware, so some tests that check that we don't do any unnecessary allocation of memory might be useful.

Some ideas:
- Have tests that checks that our supposedly memory-efficient Laplace variants (last-layer, subnetwork, KFAC, low-rank) actually scale to large models and don't throw out-of-memory errors 
- Have tests that fail when trying to be bold and e.g. run full Laplace on a big model (this one is kind of trivial)

Are there actually good ways to test these things? I.e. do we know which hardware the tests will be run on (probably on some CPU) and can we artificially limit the memory available (e.g. to emulate behaviour for a standard GPU with, say, 16GB of memory, if the CPU comes with more RAM than that)?

Somewhat relatedly, we could consider adding more informative error messages when running out of memory during Hessian allocation / computation. E.g., if initialising the Hessian runs out of memory, we could raise an error saying something like
"Your model is too big for using FullLaplace. It has X parameters, so the Hessian would be YTB large, while your CPU/GPU only has ZGB memory available. To use FullLaplace on your machine, your model can at most have ~V parameters. Instead, consider using a more memory-efficient Laplace variant, such as W."

runame:
@edaxberger I think this issue was sufficiently addressed in #72. I will create a new one for the descriptive error message.

PR #72, opened 2021-12-15
Keep initialization of H for all-weights and last-layer separate

runame:
This is an alternative to #71: it addresses a bug resulting from #62, first reported [here](https://github.com/runame/laplace-experiments/pull/17) (note: links to PR discussion in private repository). For the last-layer LA flavors the Hessian approximation `H` was first initialized for all-weights, which leads to out-of-memory errors for larger models.

The advantage of the previous fix is that it keeps the classes for all-weights and last-layer flavors more strictly separated, which might make it harder to introduce similar bugs in the future. However, many more classes are necessary. @AlexImmer and me agreed that this additional complexity is probably not worth it.

Changes:
- Add better tests for the initialization of all the Laplace classes, also with a large model (Wide ResNet 50-2), as suggested in #69.
- Fix the `H` initialization bug. The `posterior_precision` falls back to the prior before calling `fit()` for the first time for most cases. Exceptions: a last-layer flavor which doesn't get the `last_layer_name` passed as an argument and low-rank Laplace. For these two cases, `H` will be `None`. When trying to call `posterior_precision` in these cases, a descriptive error will be raised.
- Remove redundant code + minor fixes.

### DIFF
--- a/laplace/baselaplace.py
+++ b/laplace/baselaplace.py
@@ -1,5 +1,4 @@
 from math import sqrt, pi, log
-from laplace.curvature.asdl import AsdlHessian
 import numpy as np
 import torch
 from torch.nn.utils import parameters_to_vector, vector_to_parameters
@@ -7,10 +6,11 @@ from torch.distributions import MultivariateNormal, Dirichlet, Normal
 
 from laplace.utils import parameters_per_layer, invsqrt_precision, get_nll, validate
 from laplace.matrix import Kron
-from laplace.curvature import BackPackGGN, BackPackEF, AsdlGGN, AsdlEF
+from laplace.curvature import BackPackGGN, AsdlHessian
 
 
-__all__ = ['BaseLaplace', 'FullLaplace', 'KronLaplace', 'DiagLaplace', 'ParametricLaplace']
+__all__ = ['BaseLaplace', 'ParametricLaplace',
+           'FullLaplace', 'KronLaplace', 'DiagLaplace', 'LowRankLaplace']
 
 
 class BaseLaplace:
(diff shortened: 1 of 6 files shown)


edaxberger:
Looks great, and much simpler than the other solution indeed! I also tested it with WILDS and it works well.

Is it a problem that for last-layer with no `last_layer_name` passed and low-rank, `posterior_precision` is not defined? Does this mean that doing continual learning would be more difficult with these flavours (I guess even if so, we wouldn't want people to use those for CL anyways)?

edaxberger (approved):

runame:
Thanks for testing it with WILDS!

I don't think the two exceptions are a problem:
1. As you say, last-layer flavors should most likely not be used for continual learning anyway. Also, it is still possible to use them by passing the `last_layer_name` argument or simply a few more lines of code in the actual continual learning script.
2. Low-rank does not even support fitting repeatedly (without overriding `H`), hence it is not an option for continual learning anyway.

edaxberger:
Yes, good points!

AlexImmer:
We could additionally prohibit trying to do CL with these classes by adjusting the `fit` method to have no `override` argument and default to `override=True`.

runame:
I think that's also ok. I don't really have any use case in mind where one might want to use `override=False` with the last-layer flavors. And if that changes, we can easily enable the option again. Alternatively, we can raise an error like we currently do for low-rank Laplace, to avoid confusion of the user (in principle there is no reason why there should be no `override` argument for last-layer flavors).

runame:
Now a descriptive error gets raised when `override=False` for low-rank or last-layer Laplace approximations.

edaxberger:
Great, I agree that a descriptive error is more useful/clear than just not offering the option at all (and we might still add the feature at some point if we think it's useful at all). Happy to merge this in.

runame:
I think @AlexImmer wanted to take a closer look today. After that we can merge it.

AlexImmer:
lgtm

AlexImmer (approved):

PR #74, opened 2021-12-22
Fix device bug in eig_lowrank

### LINKED ISSUE #73: AsdlHessian data and model on separate devices
Hi guys,
I have been playing around with this library(it's really good!).
This is just a small thing but when losses are calculated in the [eig_lowrank](https://github.com/AlexImmer/Laplace/blob/7e42de812032cbf5d1214b89fe9cbd181c461942/laplace/curvature/asdl.py#L158)
in asdl.py I think the data and the model are not on the same device because a train_loader is passed to the eig_lowrank function. A simple change - .to(device) would make it work. 

Keep up the good work! :)

runame:
Fixes #73.


AlexImmer (approved):

AlexImmer:
Thanks for fixing this.

Issue #86, opened 2022-01-26
Questions about Subnetwork


FrederikWarburg:
Hi!

Thanks for the cool new feature about the subnetwork. I have some comments and questions.

1) Bug in check for subnetwork

Running on GPU, it fails as you have a check for torch.longtensor, but the indices will be of type torch.cuda.longtensor and the program fails. Possible fix in utils/subnetwork line 94 change to:

```
        elif not ((isinstance(subnetwork_indices, torch.LongTensor) or isinstance(subnetwork_indices, torch.cuda.LongTensor)) and
            subnetwork_indices.numel() > 0 and len(subnetwork_indices.shape) == 1):
            raise ValueError('Subnetwork indices must be non-empty 1-dimensional torch.LongTensor.')
```

2) Hyper optimizer

In your standard example, you first have trainer.fit() followed by hyperparameter optimisation like this:

```
    log_prior, log_sigma = torch.ones(1, requires_grad=True), torch.ones(1, requires_grad=True)
    hyper_optimizer = torch.optim.Adam([log_prior, log_sigma], lr=1e-1)
    for i in range(n_epochs):
        hyper_optimizer.zero_grad()
        neg_marglik = - la.log_marginal_likelihood(log_prior.exp(), log_sigma.exp())
        neg_marglik.backward()
        hyper_optimizer.step()
    
    print(f"sigma={la.sigma_noise.item():.2f}")
    print(f"prior precision={la.prior_precision.item():.2f}")
```

Can you provide an example of how you would do something similar with the subnetwork? When I try naively, I get a dimension error:

```
return (delta * self.prior_precision_diag) @ delta
RuntimeError: The size of tensor a (2701) must match the size of tensor b (100) at non-singleton dimension 0
```

Let me know if I should provide more details on network etc.

3) DiagLaplace for subnetwork

I would like to use the diagonal hessian structure for subnetwork. Could you provide me with some pointers to how I would do this? If I understand correctly, I cannot just:

```
import torch
subnetwork_indices = torch.tensor([diagonal elements])

```
as this will also account for the correlation between the diagonal elements. What would be the best way to combine subnetwork and laplace.DiagLaplace?

4) Strange warning:

If I code my network like this:

```
class Model(nn.Module):
            def __init__(self):
                super(Model, self).__init__()
                self.l1 = torch.nn.Linear(1, 50)
                self.l2 = torch.nn.Linear(50, 100)
                self.l3 = torch.nn.Linear(100, 50)
[...]

PR #87, opened 2022-01-31
Fixes and features for SubnetLaplace

edaxberger:
This PR aims to address #86. Namely, this implements the following changes:
- It fixes an issue where the subnet indices validity check would fail on GPU as we only allowed `torch.LongTensors` but not `torch.cuda.LongTensors`.
- It technically enables marginal likelihood computation and therefore optimization with `SubnetLaplace` (i.e. essentially makes sure that there are no dimension errors when calling `la.log_marginal_likelihood()` on a `SubnetLaplace` model); however we should generally not recommend using this, as the marginal likelihood is probably not meaningful when computed over a subnetwork (without any additional care that would likely require original research); so I'm not sure whether we should have this feature if we don't know how it'll behave.
- It adds `DiagSubnetLaplace`, a subclass of `SubnetLaplace` that uses a diagonal Hessian approximation (the existing full Hessian variant is now called `FullSubnetLaplace`); we initially thought that this wouldn't be useful as a full diagonal LA would probably be better in most cases and not incur much overhead, but this feature was explicitly requested in #86, so why not.

Let me know what you think!

### DIFF
--- a/laplace/subnetlaplace.py
+++ b/laplace/subnetlaplace.py
@@ -159,6 +159,15 @@ class DiagSubnetLaplace(SubnetLaplace, DiagLaplace):
     def _init_H(self):
         self.H = torch.zeros(self.n_params_subnet, device=self._device)
 
+    def _check_jacobians(self, Js):
+        if not isinstance(Js, torch.Tensor):
+            raise ValueError('Jacobians have to be torch.Tensor.')
+        if not Js.device == self._device:
+            raise ValueError('Jacobians need to be on the same device as Laplace.')
+        m, k, p = Js.size()
+        if p != self.n_params_subnet:
+            raise ValueError('Invalid Jacobians shape for Laplace posterior approx.')
+
     def sample(self, n_samples=100):
         # sample only subnetwork parameters and set all other parameters to their MAP estimates
         samples = torch.randn(n_samples, self.n_params_subnet, device=self._device)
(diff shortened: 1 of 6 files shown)


laplace/subnetlaplace.py
```diff
@@ -117,14 +111,66 @@ def prior_precision_diag(self):
 
         else:
             raise ValueError('Mismatch of prior and model. Diagonal or scalar prior.')
+    
+    @property
+    def mean_subnet(self):
```
runame, line 116: One issue that was not introduced in this PR, but I just thought about it: the `mean` property should maybe be `mean_subset`, since "mean" implies a distribution and, by definition, in subnet LA we only have a distribution over the subset. Hence, "mean" is only meaningful in the context of the subset. I know this is mostly semantics, but could actually confuse the user, e.g. when calling `subnet_la.mean` and `subnet_la.posterior_covariance` the dimensions won't match.

runame (commented):
Sorry for the late review, looks good to me! I have raised one general concern in my comment.

laplace/subnetlaplace.py
edaxberger: Thanks for looking over this, Runa! Yes, this is a good point. I've also thought about this, and have considered the solution you proposed, but that would have made some parts of the subnet laplace implementation more complex (can't remember what exactly). But I agree that it makes more sense conceptually and am happy to change this -- I'll open a separate issue for this!

edaxberger:
@AlexImmer Feel free to take a look once you have some time (no rush of course). Would be great to merge this in at some point, as there seems to be some interest in this, now again with issue #90. Thanks a lot!

AlexImmer:
Sorry for holding this back. looks good to me.

Issue #93, opened 2022-03-28
predictive samples GLM


georgezefko:
Hey,

Thank you for this package.

I am using the library for one of my projects. I noticed when I call the la(input,pred_type='glm') everything is smooth.
However, when i call la.predictive_samples(input,pred_type='glm')  i get the following value error due to MultivariateNormal function in line **593**. 

In lines **539-543** you have created a try-except statement that uses **Normal**  instead **MultivariateNormal**.
If i switch to Normal in line **593** the predictive_sample function runs smoothly.

Do you know if this is some sort of bug or am I doing something wrong?
Additionally, is it correct to use Normal instead MultivariateNormal in line 593?

`ValueError                                Traceback (most recent call last)
/Users/georgioszefkilis/Bayesian_Deep_Learning/notebooks/Laplace_STNseq_misMNIST.ipynb Cell 28' in <module>
      [1](vscode-notebook-cell:/Users/georgioszefkilis/Bayesian_Deep_Learning/notebooks/Laplace_STNseq_misMNIST.ipynb#ch0000029?line=0)[ get_ipython().run_line_magic('load_ext', 'autoreload')
      ]()[2](vscode-notebook-cell:/Users/georgioszefkilis/Bayesian_Deep_Learning/notebooks/Laplace_STNseq_misMNIST.ipynb#ch0000029?line=1)[ get_ipython().run_line_magic('autoreload', '2')
----> ]()[3](vscode-notebook-cell:/Users/georgioszefkilis/Bayesian_Deep_Learning/notebooks/Laplace_STNseq_misMNIST.ipynb#ch0000029?line=2)[ preds= sub_la.predictive_samples(test,pred_type='glm',n_samples=100)

File ~/Laplace/laplace/baselaplace.py:593, in ParametricLaplace.predictive_samples(self, x, pred_type, n_samples, images)
    ]()[591](file:///Users/georgioszefkilis/Laplace/laplace/baselaplace.py?line=590)[ assert f_var.shape == torch.Size([f_mu.shape[0], f_mu.shape[1], f_mu.shape[1]])
    ]()[592](file:///Users/georgioszefkilis/Laplace/laplace/baselaplace.py?line=591)[ #print('fvar shape',f_var.shape)
--> ]()[593](file:///Users/georgioszefkilis/Laplace/laplace/baselaplace.py?line=592)[ dist = MultivariateNormal(f_mu, f_var)
[...]

Issue #91, opened 2022-03-15
Smooth samples


philipphennig:
Hi all (not sure if you’re all still hanging out on slack),

I’ve just had a bit of fun playing with `laplace-torch` again to make a quick plot. One thing that keeps being a bit of a nuisance is that predictive plots with `link_approx="mc"` look “frayed”. I realised just now that this is because you're using different samples for each input location. Of course, the plot would be smooth if the samples were all the same for each input. 

This seems to be down to the code around [line 540 in baselaplace.py](https://github.com/AlexImmer/Laplace/blob/9a37ed7445d485e4c4c13900d8dd509976d7b55e/laplace/baselaplace.py#L539)

which calls a `torch.dist.normal.sample` . Wouldn’t it be better to draw just `n_samples` standard Gaussian RVs here and then broadcast those same samples to all locations in the input, multiplying with the sqrt-covariance and adding the local mean? This would have three advantages:

* it should be faster (no need to draw an insane number of RVs)
* the predictive then inherits the smoothness of the underlying logit layer. Can thus be used as an objective for all sorts of fun
* for the same reason, plots look nicer :) 

Come to think of it, it would be even nicer if those samples could be actively triggered, so people can keep them constant if they want to optimize (for active learning)

What do you think? 

Best,

Philipp

AlexImmer:
Hi Philipp, thanks for the great suggestion. I agree with all the points you made and implemented the corresponding changes. The plots definitely look much nicer and it should be significantly faster. In case you want to control the seed of the samples, you will be able to pass a `torch.Generator` to the sampling and predictive function.

philipphennig:
Awesome, thanks! I'll definitely try it out once this is merged into `main`. Thanks for taking care of this!

PR #95, opened 2022-04-04
Improve predictive samples

AlexImmer:
Addresses #93 and #91 with following changes:
- add function to sample normal using broadcasting (#91) which leads to smoother predictives
- allow controlling randomness by passing `torch.Generator` (#91)
- diagonalize posterior predictive on output level only when asked for explicitly (#93)

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,5 +1,5 @@
 import torch
-from laplace.utils import invsqrt_precision, diagonal_add_scalar, symeig
+from laplace.utils import invsqrt_precision, diagonal_add_scalar, symeig, normal_samples
 
 
 def test_sqrt_precision():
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -33,4 +33,31 @@ def test_symeig_custom_low_rank():
     # symeig should fail for low-rank
     assert not torch.all(l1 >= 0.0)
     # test clamping to zeros
-    assert torch.all(l2 >= 0.0)
\ No newline at end of file
+    assert torch.all(l2 >= 0.0)
+
+    
+def test_diagonal_normal_samples():
```
runame, line 39: Is it possible to test against torch `Normal`/`MultivariateNormal` samples by using the same generator state or is the implementation not the same?

runame (approved):
Looks good to me! See comments below for nit questions.

tests/test_utils.py
AlexImmer: Checked it, unfortunately not. But the statistics match when taking many samples.

Issue #94, opened 2022-04-01
Use ASDL as the default for classification


wiseodd:

edaxberger:
Addressed in #96.

PR #96, opened 2022-04-11
Use ASDL as the default for classification

edaxberger:
This PR addresses #94 by simply setting `backend=None` as the default `Laplace` constructor argument and then determining the backend to use at runtime depending on the likelihood, i.e.
```python3
if backend is None:
    backend = AsdlGGN if likelihood == 'classification' else BackPackGGN
```

Let me know what you think!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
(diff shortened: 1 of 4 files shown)


AlexImmer (approved):
Makes sense, looks good to me.

PR #117, opened 2022-11-08
Fix device and dtype of `block_diag` used for `Kron.to_matrix()`

runame:
Before this fix calling `to_matrix()` on an instance of the `Kron` class did not preserve the dtype and device, so even when everything is computed on GPU the returned matrix used to be on CPU, which is unexpected behavior.


AlexImmer (approved):
👍🏻

PR #120, opened 2023-03-17
Enable BayesOpt

wiseodd:
Changes:

* Remove `detach()` in GLM predictive so that one can backprop from `f(x)` to `x`, in particular for optimizing acquisition functions.
* Add `functional_covariance` method for GLM predictive that returns the `(nk,)` mean and the `(nk, nk)` covariance over `n` input points and `k` outputs.
* Add example implementation of a linearized-Laplace-based BoTorch model for BayesOpt.

### DIFF
--- a/examples/bayesopt_example.py
+++ b/examples/bayesopt_example.py
@@ -1,0 +1,313 @@
+import warnings
+warnings.filterwarnings('ignore')
+
+import numpy as np
+import functorch as ft
+import torch
+from torch import nn, optim
+from torch import distributions as dists
+from torch.nn import functional as F
+from gpytorch import distributions as gdists
+import torch.utils.data as data_utils
+from botorch.models.model import Model
+from botorch.posteriors.gpytorch import GPyTorchPosterior 
+import tqdm
+
+from laplace import Laplace
+from laplace.curvature import AsdlGGN, BackPackGGN, AsdlHessian
+
+import argparse, sys
+
+from botorch.models.gp_regression import SingleTaskGP
+from botorch.test_functions import Ackley, Branin
+from botorch.acquisition.analytic import ExpectedImprovement, UpperConfidenceBound
+from botorch.acquisition.monte_carlo import qExpectedImprovement
+from botorch.optim.optimize import optimize_acqf
+
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Hashable,
+    List,
+    Mapping,
+    Optional,
+    TYPE_CHECKING,
+    TypeVar,
+    Union,
+)
+
+
+class LaplaceBNN(Model):
+    """
+    Install first:
+    pip install laplace-torch
+    """
+
+    def __init__(
+            self,
+            train_X: torch.Tensor,
+            train_Y: torch.Tensor,
+            bnn: Laplace = None,
+            likelihood: str = 'regression',
+            batch_size: int = 1024):
+
+        super().__init__()
+
+        self.train_X = train_X
+        self.train_Y = train_Y
+        self.likelihood = likelihood
+        self.batch_size = batch_size
+        self.nn = nn.Sequential(
+            nn.Linear(train_X.shape[-1], 50),
+            nn.ReLU(),
+            nn.Linear(50, 50),
+            nn.Tanh(),
+            nn.Linear(50, train_Y.shape[-1])
+        )
+        self.bnn = bnn
+
+        if self.bnn is None:
+            self._train_model(self._get_train_loader())
+
+
+    def posterior(
+        self,
+        X: torch.Tensor,
+        output_indices: Optional[List[int]] = None,
+        observation_noise: bool = False,
+        posterior_transform=None,
+        **kwargs: Any,
+    ) -> GPyTorchPosterior:
+        """
+        Note: q is the num. of x's predicted jointly
+        """
+        # Transform to `(batch_shape*q, d)`
+        B, Q, D = X.shape
+        X = X.reshape(B*Q, D)
+
+        # Posterior predictive distribution
+        # mean_y is (batch_shape*q, k); cov_y is (batch_shape*q*k, batch_shape*q*k)
+        mean_y, cov_y = self._get_prediction(X, joint=True)
+
+        # Mean in `(batch_shape, q*k)`
+        K = self.num_outputs
+        mean_y = mean_y.reshape(B, Q*K)
+
+        # Cov is `(batch_shape, q*k, q*k)`
+        cov_y += 1e-4*torch.eye(B*Q*K)
+        cov_y = cov_y.reshape(B, Q, K, B, Q, K)
+        cov_y = torch.einsum('bqkbrl->bqkrl', cov_y)  # (B, Q, K, Q, K)
+        cov_y = cov_y.reshape(B, Q*K, Q*K)
+
+        dist = gdists.MultivariateNormal(mean_y, covariance_matrix=cov_y)
+        post_pred = GPyTorchPosterior(dist)
+
+        if posterior_transform is not None:
+            return posterior_transform(post_pred)
+
+        return post_pred
+
+
+    def condition_on_observations(self, X: torch.Tensor, Y: torch.Tensor, **kwargs: Any) -> Model:
+        self.train_X = torch.cat([self.train_X, X], dim=0)
+        self.train_Y = torch.cat([self.train_Y, Y], dim=0)
+
+        train_loader = self._get_train_loader()
+        self._train_model(train_loader)
+
+        return LaplaceBNN(
+            # Added dataset & retrained BNN
+            self.train_X, self.train_Y, self.bnn,
+            self.likelihood, self.batch_size
+        )
+
+    @property
+    def num_outputs(self) -> int:
+        r"""The number of outputs of the model."""
+        return self.train_Y.shape[-1]
+
+    def _train_model(self, train_loader):
+        """
+        Train BNN with the Laplace approximation
+        """
+        n_epochs = 1000
+        optimizer = optim.Adam(self.nn.parameters(), lr=1e-1, weight_decay=1e-3)
+        scheduler = optim.lr_scheduler.CosineAnnealingLR(optimizer, n_epochs*len(train_loader))
+        loss_func = nn.MSELoss()
+
+        for i in range(n_epochs):
+            for x, y in train_loader:
+                optimizer.zero_grad()
+                output = self.nn(x)
+                loss = loss_func(output, y)
+                loss.backward()
+                optimizer.step()
+                scheduler.step()
+
+        self.nn.eval()
+
+        self.bnn = Laplace(
+            self.nn, self.likelihood,
+            subset_of_weights='all', hessian_structure='kron',
+        )
+        self.bnn.fit(train_loader)
+        self.bnn.optimize_prior_precision(n_steps=30)
+
+
+    def _get_prediction(self, test_x: torch.Tensor, joint=True, use_test_loader=False):
+        """
+        Batched Laplace prediction.
+
+        Args:
+            test_x: Tensor of size `(batch_shape, d)`.
+
+        Returns:
+            Tensor of size `(batch_shape, k)`
+        """
+        if self.bnn is None:
+            print('Train your model first before making prediction!')
+
+        if not use_test_loader:
+            mean_y, cov_y = self.bnn(test_x, joint=joint)
+        else:
+            test_loader = data_utils.DataLoader(
+                data_utils.TensorDataset(test_x, torch.zeros_like(test_x)),
+                batch_size=256
+            )
+
+            mean_y, cov_y = [], []
+
+            for x_batch, _ in test_loader:
+                _mean_y, _cov_y = self.bnn(x_batch, joint=joint)
+                mean_y.append(_mean_y)
+                cov_y.append(_cov_y)
+
+            mean_y = torch.cat(mean_y, dim=0).squeeze()
+            cov_y = torch.cat(cov_y, dim=0).squeeze()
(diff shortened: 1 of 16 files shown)


AlexImmer:
Looks good to me. A few comments/questions:
- Could you add a small test case where the predictives are currently tested but with the setting `joint=True` asserting 1) equal marginal predictive and 2) right shape?
- To differentiate the predictive uncertainty as well, which depends on Jacobians, the Jacobians would have to be differentiable. Don't you need that? This would require `create_graph=True` apart from `retain_graph=True`.
- How well does the Laplace model work vs. the GP on these toy problems?

laplace/baselaplace.py
```diff
@@ -628,10 +636,16 @@ def predictive_samples(self, x, pred_type='glm', n_samples=100,
...
 
     @torch.enable_grad()
-    def _glm_predictive_distribution(self, X):
+    def _glm_predictive_distribution(self, X, joint=False):
         Js, f_mu = self.backend.jacobians(X)
-        f_var = self.functional_variance(Js)
-        return f_mu.detach(), f_var.detach()
+
+        if joint:
+            f_mu = f_mu.flatten()  # (batch*out)
```
AlexImmer, line 643: Could this be risky when we have multi-output regression or do we not support this anyway?

laplace/curvature/backpack.py
```diff
@@ -41,7 +41,7 @@ def jacobians(self, x):
                 if model.output_size > 1:
                     out[:, i].sum().backward()
                 else:
-                    out.sum().backward()
+                    out.sum().backward(retain_graph=True)
```
AlexImmer, line 44: Here potentially `create_graph=True` if pred uct. needs to be differentiable. We should, at least in the future, consider this to be a backend option to turn on and off since it costs more memory to have these set to true.

AlexImmer (approved):

examples/bayesopt_example.py
```diff
@@ -0,0 +1,313 @@
+import warnings
+warnings.filterwarnings('ignore')
+
+import numpy as np
+import functorch as ft
+import torch
+from torch import nn, optim
+from torch import distributions as dists
+from torch.nn import functional as F
+from gpytorch import distributions as gdists
```
runame, line 10: Since there are multiple packages used here which we don't require for the laplace package itself, we might want to add an additional `requirements.txt` file or similar for the example scripts.

examples/regression_example.py
```diff
@@ -47,9 +47,22 @@ def get_model():
...
+# 1.) Marginal predictive distribution N(f_map(x_i), var(x_i))
+# The mean is (m,k), the var is (m,k,k)
+f_mu, f_var = la(X_test)  
+
+# 2.) Joint pred. dist. N((f_map(x_1),...,f_map(x_m)), Cov(f(x_1),...,f(x_m)))
+# The mean is (m*k,) where k is the output dim. The cov is (m*k,m*k)
+f_mu_joint, f_cov = la(X_test, joint=True)  
+
+# Both should be true
+print(torch.allclose(f_mu.flatten(), f_mu_joint))
```
runame, line 61: I think using `assert` here is nicer.

setup.cfg
```diff
@@ -38,7 +38,7 @@ install_requires =
     backpack-for-pytorch
     asdfghjkl
 # Require a specific Python version, e.g. Python 2.7 or >= 3.4
-python_requires = >=3.8
+python_requires = >=3.7
```
runame, line 41: Any particular reason for the downgrade?

laplace/curvature/asdl.py
```diff
@@ -63,7 +63,7 @@ def gradients(self, x, y):
         Gs : torch.Tensor
             gradients `(batch, parameters)`
         """
-        f = batch_gradient(self.model, self.lossfunc, x, y).detach()
+        f = batch_gradient(self.model, self.lossfunc, x, y)
```
runame, line 66: Does the `detach()` also have to be removed from for the Jacobians [here](https://github.com/AlexImmer/Laplace/blob/main/laplace/curvature/asdl.py#L42)?

laplace/curvature/backpack.py
```diff
@@ -41,7 +41,7 @@ def jacobians(self, x):
...
                 to_cat = []
                 for param in model.parameters():
                     to_cat.append(param.grad_batch.detach().reshape(x.shape[0], -1))
```
runame, line 47: This `detach()` is fine?

runame (approved):
LGTM!

Are there cases where you might want to differentiate through NN samples, i.e. [here](https://github.com/AlexImmer/Laplace/blob/main/laplace/baselaplace.py#L640)? Also, what about `LowRankLaplace` and `LLLaplace`, see [here](https://github.com/AlexImmer/Laplace/blob/main/laplace/baselaplace.py#L923) and [here](https://github.com/AlexImmer/Laplace/blob/main/laplace/lllaplace.py#L114)?

laplace/curvature/backpack.py
wiseodd: Fixed [bc50025](https://github.com/AlexImmer/Laplace/pull/120/commits/bc50025d079abae01a65ed1dce8c3e3fe0975e6b)

setup.cfg
wiseodd: Many people I talked to still have TF v1.x as a dependency for various reasons. The last TF v1.x doesn't support Py > 3.7.

laplace/curvature/backpack.py
wiseodd: Fixed in [bc50025](https://github.com/AlexImmer/Laplace/pull/120/commits/bc50025d079abae01a65ed1dce8c3e3fe0975e6b)

laplace/curvature/asdl.py
wiseodd: Fixed in [6e50ab4](https://github.com/AlexImmer/Laplace/pull/120/commits/6e50ab4e774b2834b11aa5e984bdd1c069255b49)

wiseodd:
@AlexImmer 

> Could you add a small test case where the predictives are currently tested but with the setting `joint=True` asserting 1) equal marginal predictive and 2) right shape?
> * To differentiate the predictive uncertainty as well, which depends on Jacobians, the Jacobians would have to be differentiable. Don't you need that? This would require `create_graph=True` apart from `retain_graph=True`.

Done in [bc50025](https://github.com/AlexImmer/Laplace/pull/120/commits/bc50025d079abae01a65ed1dce8c3e3fe0975e6b).

> * How well does the Laplace model work vs. the GP on these toy problems?

Seems to work reasonably:

```
agustinuskristiadi@akmbp ~/Projects/Laplace                                                               [13:37:28]
(base) > $ python examples/bayesopt_example.py --exp_len 50 --test_func branin --acqf qEI                [±detach ●]

Test Function: branin
-------------------------------------

[BNN-LA, MSE = 85.071; Best f(x) = 0.846, curr f(x) = 3.055]: 100%|██████████████████| 50/50 [01:58<00:00,  2.37s/it]
[GP, MSE = 2081.004; Best f(x) = 0.510, curr f(x) = 12.377]: 100%|███████████████████| 50/50 [00:13<00:00,  3.83it/s]
                                                                                                                     
agustinuskristiadi@akmbp ~/Projects/Laplace                                                               [13:39:57]
(base) > $ python examples/bayesopt_example.py --exp_len 50 --test_func branin --acqf EI                 [±detach ●]

Test Function: branin
-------------------------------------

[BNN-LA, MSE = 183.844; Best f(x) = 0.939, curr f(x) = 125.097]: 100%|███████████████| 50/50 [00:59<00:00,  1.19s/it]
[GP, MSE = 4422.465; Best f(x) = 1.199, curr f(x) = 2.495]: 100%|████████████████████| 50/50 [00:06<00:00,  8.30it/s]
```

@runame 

> Are there cases where you might want to differentiate through NN samples? 

There is, e.g. in Thompson sampling like https://arxiv.org/abs/1706.01825. I added the option.

> Also, what about LowRankLaplace

There's already `functional_covariance` function for `LowRankLaplace` here https://github.com/AlexImmer/Laplace/blob/6e50ab4e774b2834b11aa5e984bdd1c069255b49/laplace/baselaplace.py#L1016. Do I need to do something else?

> LLLaplace?

[f0981a7](https://github.com/AlexImmer/Laplace/pull/120/commits/f0981a733b52adb9f0a52e322102c935a6b0765a)

wiseodd:
@AlexImmer, @runame: For making the last-layer Jacobians optionally backpropable, I need to do the following. I'm not sure how to make `forward_with_features` communicate with the hook, though. 

```python
def forward_with_features(self, x: torch.Tensor, enable_backprop=True)
    """
    Add an option `enable_backprop`
    """
    pass

def _get_hook(self, name: str) -> Callable:
    def hook(_, input, __):
        # Depending on the above `enable_backprop`, we detach
        self._features[name] = input[0].detach()
    return hook
```

Should I `enable_backprop` a class-level attribute? (In this case I also need to do the same with `BackPackInterface`.)

runame:
> > Also, what about LowRankLaplace
> 
> There's already functional_covariance function for LowRankLaplace here
> 
> [Laplace/laplace/baselaplace.py](https://github.com/AlexImmer/Laplace/blob/6e50ab4e774b2834b11aa5e984bdd1c069255b49/laplace/baselaplace.py#L1016)
> 
> Line 1016 in [6e50ab4](https://github.com/AlexImmer/Laplace/commit/6e50ab4e774b2834b11aa5e984bdd1c069255b49)
>  def functional_covariance(self, Jacs): 
> . Do I need to do something else?

I was asking because of [this line](https://github.com/AlexImmer/Laplace/blob/main/laplace/baselaplace.py#L923): you detached `self.mean` for all `ParametricLaplace` subclasses, but not for LowRankLaplace.

runame:
> For making the last-layer Jacobians optionally backpropable, I need to do the following. (...)

Yeah I think a class-level attribute is probably the best solution. Also, maybe it makes sense to consistently name the argument which decides wether to detach or not; in some places it's called `detach` and here `enable_backprop`.

laplace/lllaplace.py
```diff
@@ -113,16 +113,23 @@ def fit(self, train_loader, override=True):
         super().fit(train_loader, override=override)
         self.mean = parameters_to_vector(self.model.last_layer.parameters()).detach()
```
runame, line 114: @wiseodd Does the `detach()` have to stay here? You removed it for `ParametricLaplace` subclasses (same question as for `LowRankLaplace`).

examples/bayesopt_example.py
wiseodd: [7726b3d](https://github.com/AlexImmer/Laplace/pull/120/commits/7726b3d9ee431e2a6c258ab27ddecd3b60d17b56)

examples/regression_example.py
wiseodd: [7726b3d](https://github.com/AlexImmer/Laplace/pull/120/commits/7726b3d9ee431e2a6c258ab27ddecd3b60d17b56)

wiseodd:
@AlexImmer, @runame how do you actually enable backprop in ASDL's jacobian [here](https://github.com/AlexImmer/Laplace/blob/ada1c6f9a4aa879939de52c8fa454ff27cbfe5a4/laplace/curvature/asdl.py#L21)? (I'm unfamiliar with its API.)

laplace/lllaplace.py
wiseodd: > Yeah I think a class-level attribute is probably the best solution. Also, maybe it makes sense to consistently name the argument which decides wether to detach or not; in some places it's called detach and here enable_backprop.

See [7726b3d](https://github.com/AlexImmer/Laplace/pull/120/commits/7726b3d9ee431e2a6c258ab27ddecd3b60d17b56)

wiseodd:
> Yeah I think a class-level attribute is probably the best solution. Also, maybe it makes sense to consistently name the argument which decides wether to detach or not; in some places it's called detach and here enable_backprop.

See [7726b3d](https://github.com/AlexImmer/Laplace/pull/120/commits/7726b3d9ee431e2a6c258ab27ddecd3b60d17b56)

laplace/lllaplace.py
wiseodd: I'll update it once I know how to enable backdrop in ASDL since `LowRankLaplace` only supports ASDL.

wiseodd:
@AlexImmer @runame: There's no more action for me in this PR, so you can review it again.
 
As for enabling all these stuffs in the ASDL backend, how about we invite Michael Aerni to contribute? If he has the code already, I can also do it---in this case, he simply needs to push his changes to a new branch. In any case, let's do this in a separate PR.

laplace/baselaplace.py
```diff
@@ -522,7 +538,7 @@ def __call__(self, x, pred_type='glm', link_approx='probit', n_samples=100,
             Only works for `pred_type='glm'` and `link_approx='mc'`.
 
         generator : torch.Generator, optional
-            random number generator to control the samples (if sampling used)
+            random number generator to control the samples (if sampling used)=
```
runame, line 541: Typo?

laplace/lllaplace.py
```diff
@@ -175,11 +192,11 @@ class KronLLLaplace(LLLaplace, KronLaplace):
     _key = ('last_layer', 'kron')
 
     def __init__(self, model, likelihood, sigma_noise=1., prior_precision=1.,
-                 prior_mean=0., temperature=1., backend=None, last_layer_name=None,
+                 prior_mean=0., temperature=1., enable_backprop=True, backend=None, last_layer_name=None,
```
runame, line 195: Why does `enable_backprop` defaults to `True` here?

runame (approved):
Let's just merge this for now.

PR #123, opened 2023-04-14
Add `prior_structure` argument to `optimize_prior_precision`

runame:
Changes:
- Add a helper function `fix_prior_prec_structure` for marglik optimization.
- Add `prior_structure` argument to `optimize_prior_precision`, such that we can easily optimize a layer-wise or diagonal prior precision post-hoc.
- Rename the method `CV` for `optimize_prior_precision` to `gridsearch`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ for a regression (MSELoss) loss function.
 In the following example, a pre-trained model is loaded,
 then the Laplace approximation is fit to the training data
 (using a diagonal Hessian approximation over all parameters),
-and the prior precision is optimized with cross-validation `'CV'`.
+and the prior precision is optimized with cross-validation `'gridsearch'`.
 After that, the resulting LA is used for prediction with
 the `'probit'` predictive for classification.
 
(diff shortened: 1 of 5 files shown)


AlexImmer:
Should we set the grid search as default and/or put a warning for post hoc prior precision optimization since the method has not been thoroughly evaluated, yet, and can lead to issues apparently?

runame:
> Should we set the grid search as default and/or put a warning for post hoc prior precision optimization since the method has not been thoroughly evaluated, yet, and can lead to issues apparently?

Yeah, maybe setting `gridsearch` as the default makes sense. A warning is a bit annoying since we cannot easily detect if someone is using `optimize_prior_precision` for marglik optimization post-hoc or online.

wiseodd:
> Rename the method `CV' for `optimize_prior_precision' to `gridsearch'.

A possible downside is that people have accustomed to `CV`. Since `optimize_prior_precision` is a public-facing method and is almost always called by users, I think it's a bit dangerous to rename it. At the very least we should support both.

PR #134, opened 2023-08-02
Add `asdl_fisher_kwargs` argument

runame:
Add `asdl_fisher_kwargs` argument to be able to specify ASDL's `FisherConfig` values like `kfac_linear`, `diag_A/B`, and `autocast`.

@AlexImmer Can you check that my changes in line 848 and 852 of `baselaplace.py` don't break anything? Not sure why this was differently implemented here than in other places.

~Also, I have noticed that the tests are broken, which is really weird since I'm almost certain that I ran them successfully for my last PR. I went back to [this commit](https://github.com/AlexImmer/Laplace/commit/99278beb81d613bccac934b11158e7066ec8d34d) and they are already broken at this point. We will have to look into this.~

~_Edit:_ Ok, just checked and this PR seems to break a few more tests than the ones that were broken already, so maybe do not merge yet, will fix this tomorrow.~

### DIFF
--- a/laplace/baselaplace.py
+++ b/laplace/baselaplace.py
@@ -689,7 +689,7 @@ class ParametricLaplace(BaseLaplace):
         for sample in self.sample(n_samples):
             vector_to_parameters(sample, self.params)
             logits = self.model(X.to(self._device), **model_kwargs).detach()
-            py += 1/n_samples * torch.softmax(logits, dim=-1)
+            py += torch.softmax(logits, dim=-1) / n_samples
         vector_to_parameters(self.mean, self.params)
         return py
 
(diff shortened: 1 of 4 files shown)


runame:
Everything fixed.

wiseodd:
Would it make sense to just include `asdl_kwargs` as part of `backend_kwargs`?

runame:
I think it is better to keep them separate; otherwise we would have to add all potential `FisherConfig` args to the backend signatures or extract the args somehow (see [this line](https://github.com/AlexImmer/Laplace/blob/main/laplace/baselaplace.py#L78)). But I'm happy to be convinced otherwise if you have something concrete in mind.

wiseodd:
Then LGTM! @AlexImmer all yours now

AlexImmer:
LGTM. I think we eventually have to clean it up a little though before we bring it into main. For example, asdl config should be part of the backend config instead of being passed to Laplace, which has to abstract away from the backends.

PR #136, opened 2023-08-11
Add support for diagonal Kronecker factors in `Kron` matrix class

runame:
To support diagonal Kronecker factors in this library we have to adopt the methods of the `Kron` matrix class to be able to handle this correctly.

### DIFF
--- a/laplace/utils/matrix.py
+++ b/laplace/utils/matrix.py
@@ -109,7 +109,14 @@ class Kron:
         for F in self.kfacs:
             Qs, ls = list(), list()
             for Hi in F:
-                l, Q = symeig(Hi)
+                if Hi.ndim > 1:
+                    # Dense Kronecker factor.
+                    l, Q = symeig(Hi)
+                else:
+                    # Diagonal Kronecker factor.
+                    l, indices = Hi.sort()
+                    # This might be too memory intensive since len(Hi) can be large.
+                    Q = torch.eye(len(Hi), dtype=Hi.dtype, device=Hi.device)[indices].T
                 Qs.append(Q)
                 ls.append(l)
             eigvecs.append(Qs)
@@ -140,14 +147,16 @@ class Kron:
                 Q = Fs[0]
                 p = len(Q)
                 W_p = W[:, cur_p:cur_p+p].T
-                SW.append((Q @ W_p).T)
+                SW.append((Q @ W_p).T if Q.ndim > 1 else (Q.view(-1, 1) * W_p).T)
                 cur_p += p
             elif len(Fs) == 2:
                 Q, H = Fs
                 p_in, p_out = len(Q), len(H)
                 p = p_in * p_out
                 W_p = W[:, cur_p:cur_p+p].reshape(B * K, p_in, p_out)
-                SW.append((Q @ W_p @ H.T).reshape(B * K, p_in * p_out))
+                QW_p= Q @ W_p if Q.ndim > 1 else Q.view(-1, 1) * W_p
+                QW_pHt = QW_p @ H.T if H.ndim > 1 else QW_p * H.view(1, -1)
+                SW.append(QW_pHt.reshape(B * K, p_in * p_out))
                 cur_p += p
             else:
                 raise AttributeError('Shape mismatch')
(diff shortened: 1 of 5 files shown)


laplace/utils/matrix.py
```diff
@@ -118,7 +109,14 @@ def decompose(self, damping=False):
...
             Qs, ls = list(), list()
             for Hi in F:
-                l, Q = symeig(Hi)
+                if Hi.ndim > 1:
+                    # Dense Kronecker factor.
+                    l, Q = symeig(Hi)
+                else:
+                    # Diagonal Kronecker factor.
+                    l = Hi
+                    # This might be too memory intensive since len(Hi) can be large.
```
AlexImmer, line 118: Remove this comment that is from sorting previously?

laplace/utils/matrix.py
```diff
@@ -46,18 +46,9 @@ def init_from_model(cls, model, device):
...
-                    P_in, P_out = p.size()
-                elif p.ndim > 2:
-                    P_in, P_out = p.shape[0], np.prod(p.shape[1:])
-
-                kfacs.append([
-                    torch.zeros(P_in, P_in, device=device),
-                    torch.zeros(P_out, P_out, device=device)
-                ])
+                kfacs.append([0.])
+            elif 4 >= p.ndim >= 2:  # fully connected or or embedding or conv
```
AlexImmer, line 50: `or or`

laplace/utils/matrix.py
```diff
@@ -204,11 +204,12 @@ def logdet(self) -> torch.Tensor:
         logdet = 0
         for F in self.kfacs:
             if len(F) == 1:
-                logdet += F[0].logdet()
+                logdet += F[0].logdet() if F[0].ndim > 1 else F[0].prod().log()
```
AlexImmer, line 207: `.log().sum()` is more stable

tests/test_baselaplace.py
```diff
@@ -63,11 +63,11 @@ def test_laplace_init(laplace, model):
     elif laplace == LowRankLaplace:
         assert lap.H is None
     else:
-        H = [[k.clone() for k in kfac] for kfac in lap.H.kfacs]
+        H = [[torch.tensor(k) for k in kfac] for kfac in lap.H.kfacs]
         lap._init_H()
         for kfac1, kfac2 in zip(H, lap.H.kfacs):
             for k1, k2 in zip(kfac1, kfac2):
-                assert torch.allclose(k1, k2)
+                assert torch.allclose(k1, torch.tensor(k2))
```
AlexImmer, line 70: What is this for?

AlexImmer (commented):
LGTM, just some suggestions and questions. The `clone()` before was, if I remember correctly, important or necessary to maintain differentiability when it's needed. I am not sure whether `torch.tensor(h)` guarantees the same.

laplace/utils/matrix.py
runame: No it is due to the `torch.eye`, since the Kronecker factor is diagonal it might be too large to explicitly build a `len(H_i) x len(H_i)` matrix.

laplace/utils/matrix.py
runame: Nice catch!

tests/test_baselaplace.py
```diff
@@ -63,11 +63,11 @@ def test_laplace_init(laplace, model):
     elif laplace == LowRankLaplace:
         assert lap.H is None
     else:
-        H = [[k.clone() for k in kfac] for kfac in lap.H.kfacs]
+        H = [[torch.tensor(k) for k in kfac] for kfac in lap.H.kfacs]
```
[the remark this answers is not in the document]
runame: It is now a float scalar and not a tensor.

tests/test_baselaplace.py
runame: Same as above.

runame:
Thanks! The initialisation is now not using tensors but just float scalars, so `clone()` is not possible anymore. Not sure if this breaks something.

AlexImmer:
If it passes the tests, I guess we have to find out later.

PR #144, opened 2024-02-24
Bringing laplace-torch to foundation-model era

wiseodd:
Main features of this pull request:

1. Support only doing Laplace on params that require grad. Use case: PEFT (like LoRA) on top of a frozen foundation model. This is more efficient than SubnetLaplace since the latter still computes the full Jacobians.
2. Add support to multiple leading dims in `classification` likelihood. E.g. the logits is `(batch_size, seq_len, n_classes)`. Useful for language modeling and reward modeling.
    1. This PR also contains the `integrate-latest-asdl` changes. I tested it with my ASDL fork (only a couple of light changes to support weight-sharing dim and `ignore_index`; so crucial in language modeling): <https://github.com/wiseodd/asdl/commits/dev/>. Please also check this and let me know what's the most elegant way. 
4. Support Huggingface dataset. The assumption is that `x` is a `UserDict` containing `input_ids`, `attention_mask`, etc., things that are produced by HF dataloader.
5. Add a new likelihood called `reward_modeling` where the `classification` likelihood is used during training and the `regression` likelihood is used during prediction.
6. Add support to `torchmetrics` for gridsearch. The benefit is that it supports running metrics => less memory overhead (vis-a-vis gathering all the logits first).
7. Add Jacobian computation with `torch.func` (`functorch`, really) as a general Jacobian computation for GLM predictive. Useful for Bayesian optimization/invariance learning where you need to backprop through the variance. Much more elegant than to change ASDL. 

Relevant unit tests are provided. All tests passed; the only ones failed are the old `LowRankLaplace` issues.


edaxberger:
Thanks Agustinus for this PR, this is very useful! Just took a glance at the parts that I'm familiar with, which looks good to me. I'll defer to Alex and Runa for a more in-depth review due to their better familiarity with the core library.

laplace/utils/matrix.py
```diff
@@ -72,7 +68,7 @@ def __add__(self, other):
         if not isinstance(other, Kron):
             raise ValueError('Can only add Kron to Kron.')
 
-        kfacs = [[Hi.add(Hj) for Hi, Hj in zip(Fi, Fj)]
+        kfacs = [[Hi + Hj for Hi, Hj in zip(Fi, Fj)]
```
aleximmer, line 71: Why this change? I remember there was a reason for using `.add`, might have had to do with differentiability.

laplace/marglik_training.py
```diff
@@ -31,20 +33,22 @@ def marglik_training(
     prior_prec_init=1.,
     sigma_noise_init=1.,
     temperature=1.,
+    sigma_noise_fixed=None,
```
aleximmer, line 36: Maybe make this bool and use `sigma_noise_init` as the fixed setting for it. Otherwise, it's unclear how they overwrite each other.

laplace/utils/matrix.py
```diff
@@ -145,14 +148,16 @@ def _bmm(self, W: torch.Tensor) -> torch.Tensor:
                 Q = Fs[0]
                 p = len(Q)
                 W_p = W[:, cur_p:cur_p+p].T
-                SW.append((Q @ W_p).T)
+                SW.append((Q @ W_p).T if Q.ndim > 1 else (Q.view(-1, 1) * W_p).T)
```
aleximmer, line 151: Why the else statement although `Q = torch.eye()` for `ndim == 1`? I think in this case you could make Q above a vector of ones and save the memory?

laplace/utils/matrix.py
```diff
@@ -374,11 +384,14 @@ def _bmm(self, W: torch.Tensor, exponent: float = -1) -> torch.Tensor:
         W = W.reshape(B * K, P)
         cur_p = 0
         SW = list()
-        for ls, Qs, delta in zip(self.eigenvalues, self.eigenvectors, self.deltas):
+        for i, (ls, Qs, delta) in enumerate(zip(self.eigenvalues, self.eigenvectors, self.deltas)):
             if len(ls) == 1:
                 Q, l, p = Qs[0], ls[0], len(ls[0])
                 ldelta_exp = torch.pow(l + delta, exponent).reshape(-1, 1)
                 W_p = W[:, cur_p:cur_p+p].T
+                # print('c'); W_p.sum().backward(); input()
```
aleximmer, line 392: plaese remove debug print statements here and below

aleximmer (commented):
Thanks, looks overall good to me.

laplace/baselaplace.py
```diff
@@ -40,32 +48,55 @@ class BaseLaplace:
...
+        # Only do Laplace on params that require grad
+        self.params = []
+        self.is_subset_params = False
+        for p in model.parameters():
+            if p.requires_grad:
+                self.params.append(p)
+            else:
+                self.is_subset_params = True
+
+        self.n_params = len(parameters_to_vector(self.params).detach())
```
runame, line 72: `sum(p.numel() for p in self.params)` is a bit more efficient.

laplace/baselaplace.py
```diff
@@ -139,7 +170,7 @@ def prior_precision_diag(self):
             return self.prior_precision
 
         elif len(self.prior_precision) == self.n_layers:  # per layer
-            n_params_per_layer = parameters_per_layer(self.model)
+            n_params_per_layer = [np.prod(p.shape) for p in self.params]
```
runame, line 173: can use `p.numel()` here

laplace/baselaplace.py
```diff
@@ -580,11 +642,19 @@ def __call__(self, x, pred_type='glm', joint=False, link_approx='probit',
 
         if pred_type == 'nn' and link_approx != 'mc':
             raise ValueError('Only mc link approximation is supported for nn prediction type.')
-        
+
         if generator is not None:
             if not isinstance(generator, torch.Generator) or generator.device != x.device:
                 raise ValueError('Invalid random generator (check type and device).')
 
+        # if self.is_subset_params and pred_type == 'glm':
```
runame, line 650: Why is this commented out?

laplace/curvature/asdl.py
```diff
@@ -107,33 +128,81 @@ def _rescale_kron_factors(kron, N):
...
+        y = y if self.loss_type == LOSS_MSE else y.view(-1)
+        if 'emp' in self._ggn_type:
+            dummy = fisher_maker.setup_model_call(self._model, X)
+            dummy = dummy if self.loss_type == LOSS_MSE else dummy.view(-1, dummy.size(-1))
+            fisher_maker.setup_loss_call(self.lossfunc, dummy, y)
+        else:
+            fisher_maker.setup_model_call(self._model, X)
+        f, _ = fisher_maker.forward_and_backward()
+        # Assumes that the last dimension of f is of size outputs.
+        f = f if self.loss_type == LOSS_MSE else f.view(-1, f.size(-1))
```
runame, line 147: This is against the convention used for `CrossEntropyLoss` in PyTorch, since it expects input shape `(N, C, d_1​, d_2 ​,..., d_K​)` (see the [docs](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html)). Could lead to unexpected behaviour for users. If we go with this convention, we should comment on it in the docs and the README, as a requirement for the model.

laplace/curvature/asdl.py
```diff
@@ -107,33 +128,81 @@ def _rescale_kron_factors(kron, N):
...
```
runame, line 179: same here

laplace/curvature/asdl.py
```diff
@@ -147,19 +216,38 @@ def _ggn_type(self):
...
+        dummy = hess_maker.setup_model_call(self._model, x)
+        dummy = dummy if self.loss_type == LOSS_MSE else dummy.view(-1, dummy.size(-1))
+        y = y if self.loss_type == LOSS_MSE else y.view(-1)
+        hess_maker.setup_loss_call(self.lossfunc, dummy, y)
+        hess_maker.forward_and_backward()
         H = self._model.hessian.data
-        loss = self.lossfunc(self.model(x), y).detach()
+        f = self.model(x).detach()
+        # Assumes that the last dimension of f is of size outputs.
+        f = f if self.loss_type == LOSS_MSE else f.view(-1, f.size(-1))
```
runame, line 231: same here

laplace/curvature/asdl.py
```diff
@@ -147,19 +216,38 @@ def _ggn_type(self):
         raise NotImplementedError()
 
     def full(self, x, y, **kwargs):
-        hessian_for_loss(self.model, self.lossfunc, SHAPE_FULL, x, y)
+        if self.last_layer:
+            _, x = self.model.forward_with_features(x)
+        cfg = HessianConfig(hessian_shapes=[SHAPE_FULL])
+        hess_maker = HessianMaker(self.model, cfg)
+        dummy = hess_maker.setup_model_call(self._model, x)
+        dummy = dummy if self.loss_type == LOSS_MSE else dummy.view(-1, dummy.size(-1))
```
runame, line 224: same here

laplace/curvature/asdl.py
```diff
@@ -107,33 +128,81 @@ def _rescale_kron_factors(kron, N):
...
+    def kron(self, X, y, N, **kwargs):
+        if self.last_layer:
+            _, X = self.model.forward_with_features(X)
+        cfg = FisherConfig(fisher_type=self._ggn_type, loss_type=self.loss_type,
+                           fisher_shapes=[SHAPE_KRON], data_size=1, **kwargs)
+        fisher_maker = get_fisher_maker(self.model, cfg)
+        y = y if self.loss_type == LOSS_MSE else y.view(-1)
+        if 'emp' in self._ggn_type:
+            dummy = fisher_maker.setup_model_call(self._model, X)
+            dummy = dummy if self.loss_type == LOSS_MSE else dummy.view(-1, dummy.size(-1))
```
runame, line 173: same here

laplace/curvature/asdl.py
```diff
@@ -107,33 +128,81 @@ def _rescale_kron_factors(kron, N):
...
+        del N
+        if self.last_layer:
+            _, X = self.model.forward_with_features(X)
+        cfg = FisherConfig(fisher_type=self._ggn_type, loss_type=self.loss_type,
+                           fisher_shapes=[SHAPE_DIAG], data_size=1, **kwargs)
+        fisher_maker = get_fisher_maker(self.model, cfg)
+        y = y if self.loss_type == LOSS_MSE else y.view(-1)
+        if 'emp' in self._ggn_type:
+            dummy = fisher_maker.setup_model_call(self._model, X)
+            dummy = dummy if self.loss_type == LOSS_MSE else dummy.view(-1, dummy.size(-1))
```
runame, line 141: same here

laplace/curvature/backpack.py
```diff
@@ -127,6 +129,9 @@ def _rescale_kron_factors(kron, M, N):
     def diag(self, X, y, **kwargs):
         context = DiagGGNMC if self.stochastic else DiagGGNExact
         f = self.model(X)
+        # Assumes that the last dimension of f is of size outputs.
+        f = f if self.likelihood == 'regression' else f.view(-1, f.size(-1))
```
runame, line 133: same here

laplace/curvature/backpack.py
```diff
@@ -136,9 +141,12 @@ def diag(self, X, y, **kwargs):
 
         return self.factor * loss.detach(), self.factor * dggn
 
-    def kron(self, X, y, N, **kwargs) -> [torch.Tensor, Kron]:
+    def kron(self, X, y, N, **kwargs) -> Tuple[torch.Tensor, Kron]:
         context = KFAC if self.stochastic else KFLR
         f = self.model(X)
+        # Assumes that the last dimension of f is of size outputs.
+        f = f if self.likelihood == 'regression' else f.view(-1, f.size(-1))
```
runame, line 148: same here

laplace/curvature/backpack.py
```diff
@@ -154,6 +162,9 @@ class BackPackEF(BackPackInterface, EFInterface):
 
     def diag(self, X, y, **kwargs):
         f = self.model(X)
+        # Assumes that the last dimension of f is of size outputs.
+        f = f if self.likelihood == 'regression' else f.view(-1, f.size(-1))
```
runame, line 166: ok I will stop

laplace/curvature/curvature.py
```diff
@@ -63,8 +63,51 @@ def jacobians(self, x, enable_backprop=False):
...
+            for j, p in zip(Js.values(), params_dict.values())
+        ]
+        Js = torch.cat(Js, dim=-1)
+
+        if self.subnetwork_indices is not None:
+            Js = Js[:, :, self.subnetwork_indices]
+
+        # Compute f
+        # ------------------------
+        f = self.model(x)
```
runame, line 105: Can we avoid this additional forward pass by something like this (from the docs)?
```python
from torch.func import jacrev
x = torch.randn(5)
def f(x):
  return x.sin()
def g(x):
  result = f(x)
  return result, result
jacobian_f, f_x = jacrev(g, has_aux=True)(x)
assert torch.allclose(f_x, f(x))
```

laplace/baselaplace.py
```diff
@@ -384,31 +430,45 @@ def fit(self, train_loader, override=True):
...
-        X, _ = next(iter(train_loader))
+        data = next(iter(train_loader))
         with torch.no_grad():
-            try:
-                out = self.model(X[:1].to(self._device))
-            except (TypeError, AttributeError):
+            if isinstance(data, UserDict): # To support Huggingface dataset
+                out = self.model(data)
+            else:
+                X = data[0]
```
runame, line 453: Should there still be this here?
```python
try:
     out = self.model(X[:1].to(self._device))
except (TypeError, AttributeError):
```
It is still there for `LLLaplace`.

laplace/lllaplace.py
```diff
@@ -134,18 +142,28 @@ def _glm_predictive_distribution(self, X, joint=False):
...
         fs = torch.stack(fs)
         if self.likelihood == 'classification':
             fs = torch.softmax(fs, dim=-1)
         return fs
 
+    def _nn_predictive_classification(self, X, n_samples=100, generator=None, **model_kwargs):
+        py = 0
+        for sample in self.sample(n_samples, generator):
+            vector_to_parameters(sample, self.model.last_layer.parameters())
+            # TODO: Implement with a single forward pass until last layer.
```
runame, line 161: Maybe we should do this now, since it will make predictions a lot faster for large foundation models. Should be possible to implement with `forward_with_features()`?

laplace/baselaplace.py
```diff
@@ -228,8 +260,9 @@ def optimize_prior_precision_base(
             otherwise, the structure of init_prior_prec is maintained.
         val_loader : torch.data.utils.DataLoader, default=None
             DataLoader for the validation set; each iterate is a training batch (X, y).
-        loss : callable, default=get_nll
-            loss function to use for gridsearch.
+        loss : callable or torchmetrics.Metric, default=get_nll
```
runame, line 263: Should we change the default to the `RunningNLLMetric`?

laplace/utils/metrics.py
```diff
@@ -0,0 +1,40 @@
+import torch
+from torch.nn import functional as F
+from torchmetrics import Metric
+
+
+class RunningNLLMetric(Metric):
+    """
+    NLL metrics that
```
runame, line 8: ...

laplace/utils/metrics.py
```diff
@@ -0,0 +1,40 @@
...
+        """
+        Parameters
+        ----------
+        probs: torch.Tensor
+            probability tensor of shape (..., n_classes)
+
+        targets: torch.Tensor
+            integer tensor of shape (...)
+        """
+        probs = probs.view(-1, probs.shape[-1])
```
runame, line 31: Against the PyTorch convention again.

tests/test_baselaplace.py
```diff
@@ -406,6 +454,7 @@ def test_classification_predictive(laplace, model, class_loader):
 
     # NN predictive
     f_pred = lap(X, pred_type='nn', link_approx='mc', n_samples=100)
+    print(f_pred)
```
runame, line 457: I assume for debugging?

tests/test_curv_backends_asdl.py
```diff
@@ -77,13 +105,35 @@ def test_diag_ef_cls_kazuki_against_backpack_full(class_Xy, model):
...
+    loss += loss2
+    dggn += dggn2
+
+    # sanity check size of diag ggn
+    assert len(dggn) == model.n_params
+
+    # check against manually computed full GGN:
+    backend = BackPackEF(model2, 'regression')
+    loss_bp, dggn_bp = backend.diag(X, y)
+    print(loss, loss_bp)
```
runame, line 129: remove debugging print statements and comment here and below

tests/test_curv_backends_asdl.py
```diff
@@ -110,7 +195,21 @@ def test_kron_kazuki_vs_diag_class(class_Xy, model, Backend):
...
-def test_kron_batching_correction_kazuki(class_Xy, model, Backend):
+def test_kron_asdl_vs_diag_reg(reg_Xy, model, Backend):
+    # For a single data point, Kron is exact and should equal diag GGN
+    X, y = reg_Xy
+    backend = Backend(model, 'regression')
+    loss, dggn = backend.diag(X[:1], y[:1], N=1)
+    # sanity check size of diag ggn
+    assert len(dggn) == model.n_params
+    loss, kron = backend.kron(X[:1], y[:1], N=1)
+    print(dggn)
```
runame, line 206: here as well

runame (changes requested):
Thanks a lot for pushing this! Overall looks good to me, see my comments.

laplace/utils/matrix.py
wiseodd: Again, it's @runame change.

laplace/baselaplace.py
wiseodd: Let's leave this for a new PR

laplace/utils/metrics.py
wiseodd: As above

laplace/curvature/asdl.py
wiseodd: It's from ASDL: https://github.com/kazukiosawa/asdl/blob/9ba6a54bfa4cf3fb21dbc2922f6ef128a3b2e0f9/asdl/fisher.py#L456-L457

Does you suggest we also modify ASDL?

wiseodd: After some readings, I think the current implementation is fine. The input of the model is `N, d_1, ..., d_K, D` where `d_1, ..., d_K` are additional "batch" dimensions and `D` is the input-space dim. Then the model will always output `N, d_1, ..., d_K, C` where `C` is the number of classes.

In Huggingface, they always assume that the logits and labels are flattended, e.g. here for [GPT2](https://github.com/huggingface/transformers/blob/b109257f4fb8b1166e7c53cc5418632014ed53a5/src/transformers/models/gpt2/modeling_gpt2.py#L1336-L1338) and [LLAMA-2](https://github.com/huggingface/transformers/blob/b109257f4fb8b1166e7c53cc5418632014ed53a5/src/transformers/models/llama/modeling_llama.py#L1218-L1224). The loss is then oblivious and agnostic to the fact that we have additional batch dimensions. This is also the current assumption of Laplace and ASDL.

So, I believe as long as we consistently follow the flatten-and-compute-loss convention, we're clear. The user won't be affected at all since it's an internal implementation.

wiseodd: Looking forward to hear your thoughts, though, since I might be wrong!

laplace/lllaplace.py
wiseodd: Seems to be nontrivial at first glance. Let's do this in a separate PR.

laplace/utils/matrix.py
runame: I think this is due to changing the initialization of the Kronecker factors to not assume a specific shape by initializing with `0.0` scalars, i.e. not a tensor so `.add` would not work anymore. The motivation for this was to support diagonal Kronecker factors for large layers like embedding layers in language models. See the commit [here](https://github.com/aleximmer/Laplace/pull/144/commits/98a9800db0a28c023ba73444be3cd8d490eeca62).

runame: Not a big fan of this design, so not sure if it should be merged into `main`. Also, regarding @aleximmer's comment, are there tests for checking if the differentiability is affected?

laplace/utils/matrix.py
runame: `Q.ndim` \neq `Hi.ndim`, but I don't see any case where `Q.ndim == 1` anymore. `Q` is not a vector for diagonal Kronecker factors, because it is not necessarily the identity but some permutation matrix. The code should probably be rewritten to avoid explicitly building the permutation matrix in memory, but this requires duplicating this code in the if-else statement to deal with the diagonal Kronecker factor case separately. Btw, the PR with the changes is [here](https://github.com/aleximmer/Laplace/pull/136).

wiseodd:
Regarding the addition of `logit_class_idx` argument: I did a "smoke test" in implementing this, and realized that the changes would be substantial---even without the present PR, `laplace-torch` always assumes that `logit_class_idx == -1`. 

So, I think this is better done in a separate PR. This will also make the scope of the PRs cleaner---in this PR, we focus on HF LLM models.

wiseodd:
So, the remaining TODOs for this PR are:

1. Decide the conversations about `matrix.py`. @runame & @aleximmer could you please handle this?
2. Wait until https://github.com/f-dangel/curvlinops/pull/100 is merged

Then we can merge this PR and continue with further PRs that we deferred in this PR.

runame:
Sounds good, can you create issues for all comments that you want to address in future PRs?

wiseodd:
I replaced `get_nll` in crossval with `RunningNLLMetrics()`. So this PR will close #160 .
(I thought before that I `RunningNLLMetics` hadn't been implemented, so would have involve major work.)

README.md
```diff
@@ -22,13 +22,20 @@ The [code](https://github.com/runame/laplace-redux) to reproduce the experiments
...
+For full compatibility, install this package in a fresh virtual env.
+We assume Python >= 3.8 (last tested on Python 3.10.12).
 PyTorch version 2.0 and up is also required for full compatibility.
 To install laplace with `pip`, run the following:
 ```bash
 pip install laplace-torch
 ```
 
+
+Note that, if you would like to use ASDL as a backend, please install it via the following, for full compatibility:
```
aleximmer, line 34: Can you make more precise for what kind of compatibility? This is presumably necessary only for LLM stuff? We should try to merge these changes into ASDL if possible.

laplace/utils/matrix.py
wiseodd: I take the point of not assuming a specific shape. I think in the future this is still necessary, but let's build it from the ground up for the next release (0.3). Perhaps along with the addition of `SINGD` backend.

For the 0.2 milestone, let's revert this. (This change also breaks https://github.com/aleximmer/Laplace/issues/171 BTW)

laplace/utils/matrix.py
wiseodd: Let's leave this be for now then. We can revisit this when we add SINGD baseline. Might be useful then.

README.md
wiseodd: The PR is open. I think this is non-blocking. We can update the `setup.cfg` once Kazuki merged it.

wiseodd: https://github.com/kazukiosawa/asdl/pull/52

wiseodd:
All tasks finished!

wiseodd:
Double-checked and everything looks fine! Merging this.

Issue #138, opened 2023-11-13
Last Layer Laplace predictions could be computed much faster and becomes problematic for large label classification.


charlio23:
When experimenting with Last Layer Laplace approximation on a classification task with large labels (1K, e.g. ImageNet), the memory rapidly jumps to 10~20Gb and inference becomes slower.

For context, for a last layer 128 -> 3100 (where 3100 is the number of labels), each inference step takes ~10 seconds and 15Gb using only one sample (batch_size=1) using Diagonal Laplace. Kronecker Laplace does not fit in GPU with this setup.

When inspecting the code, I noticed the Last Layer Laplace computes the Jacobian as $\phi(x)^T \otimes I$ explicitly, which is very costly when considering >1K labels as the resulting matrix has shape (num_labels, num_features*num_labels).

**Note:** $\phi(x)$ are the features of the L-1 layer given input x, and $I$ is an identity matrix of size (num_labels, num_labels)

For the case of Last Layer Laplace, I believe this could be implemented more efficiently considering the factorisation of the jacobian. For example, posterior variance for the kronecker Laplace could be implemented as follows:

- Assume the inverse posterior precision is such that $H^{-1} = V \otimes U$, the following identity is well-established in Last Layer Laplace literature, 
$$\Sigma =  \Big(\phi(x)^T V \phi(x)\Big)U $$

Below you can find some code I used to speed up the Last Layer Laplace by modifying the code from the repository. It would be nice if you could consider making this modification as I am sure there are people interested in your package that might benefit from this.

Diagonal Laplace

```
    def _glm_predictive_distribution(self, X):
        f_mu, phi = self.model.forward_with_features(X)
        emb_size = phi.shape[-1]
        f_var torch.diag_embed(torch.matmul(self.posterior_variance.reshape(-1, emb_size),(phi*phi).transpose(0,1)).transpose(0,1))
        return f_mu.detach(), f_var.detach()

```

Kronecker Laplace
```
    def _glm_predictive_distribution(self, X):
        f_mu, phi = self.model.forward_with_features(X)

        eig_U, eig_V = self.posterior_precision.eigenvalues[0]
        vec_U, vec_V = self.posterior_precision.eigenvectors[0]
        delta = self.posterior_precision.deltas.sqrt()
        inv_U_eig, inv_V_eig = torch.pow(eig_U + delta, -1), torch.pow(eig_V + delta, -1)

[...]

PR #145, opened 2024-02-24
Add fast computation of functional_variance for DiagLLLaplace and KronLLLaplace

wiseodd:
PR for #138. Very useful for LLMs or diffusion models or any models with many outputs.

TODO: Implementation for `KronLLLaplace`. I'd like input from @aleximmer, who's the author of `matrix.py`.


laplace/lllaplace.py
```diff
@@ -210,3 +217,23 @@ class DiagLLLaplace(LLLaplace, DiagLaplace):
...
+        # each functional variance phi^2*var_weight + var_bias
+        f_var_diag_only = torch.einsum('bd,kd,bd->bk', phi, self.posterior_variance[:d*k].reshape(k, d), phi)
+
+        if self.model.last_layer.bias is not None:
+            # Add the last num_classes variances, corresponding to the biases' variances
+            # (b,k) + (1,k) = (b,k)
+            f_var_diag_only += self.posterior_variance[-k:].reshape(1, k)
+
+        # (b,k) -> (b,k,k)
+        f_var = torch.diag_embed(f_var_diag_only)
```
wiseodd, line 237: Here we might want to discuss whether this should be done or not. Currently it's done for a compatibility reason (so that it's uniform with other predictives). But imagine if we have an LLM with `num_classes = 32000`, then this is detrimental...

laplace/lllaplace.py
```diff
@@ -201,6 +208,40 @@ def __init__(self, model, likelihood, sigma_noise=1., prior_precision=1.,
     def _init_H(self):
         self.H = Kron.init_from_model(self.model.last_layer, self._device)
 
+    def _functional_variance_fast(self, X):
```
wiseodd, line 211: @aleximmer here's the initial implementation for the KronLLLaplace. The test, comparing this `f_var` to the `f_var = la.posterior_precision.inv_square_form(Js)` fails tho... 

Could you please check this? Feel free to propose a more elegant solution since you know more about the implementation of Kron.

aleximmer: What's the advantage you are trying to achieve with this? Is it faster because you do the damping formulation of the posterio update (eigenvalues + sqrt(delta))? I suppose that approximation makes the test fail potentially.

laplace/lllaplace.py
runame: I'm in favour of either not doing this or adding a flag that defaults to being consistent but inefficient, and can be set to only return the diagonal as a (batched) vector.

laplace/lllaplace.py
wiseodd: The idea is to use this identity of the matrix-Normal distribution: (See https://arxiv.org/pdf/2002.10118.pdf, Appendix B.1)

Screenshot 2024-03-02 at 5 12 51 PM

Then, it's much faster than the naive `functional_variance` since we don't need to compute the Jacobian which is `(batch_size, num_classes, num_params)`. We only need to multiply the inverse-Kronecker factors with the last layer features $\phi(x)$.

Let me know your thoughts on the best way to achieve this

wiseodd: PS. the rationale for the sqrt damping thing: I follow the KFAC-Laplace <https://openreview.net/pdf?id=Skdvd2xAZ>. Let me know what you think.

aleximmer: I looked a bit more into this now. It's a very important change to have this in since it's significantly faster. The way it should probably be implemented is by using the `damping=True/False` flag. The exact inversion with a prior would have to be done using an eigendecomposition, which we are doing right now, and can be avoided when we use the fast predictive by using the damping formulation instead. This also would avoid the recomputation of U and V from the eigendecomposition and we could add the method to `matrix.py`. I could look into how to do this best. One thing I am wondering: do you know if it is possible to do this as well for the `joint` posterior predictive?

wiseodd: Ok, great. I'll leave it to you then to do this implicitly by using damping. The corresponding test case is in `tests/test_lllaplace.py -k "test_functional_variance_fast[KronLLLaplace]"`.

wiseodd: I think joint predictive can also benefit from this ([code for naive](https://github.com/aleximmer/Laplace/blob/f14afc45094bb6e08c19800ecd9ea6a36548b827/laplace/baselaplace.py#L1006-L1012)). However, more thoughts need to be put here. So let's do this in a separate PR.

PR #146, opened 2024-03-02
Add Curvlinops backend & add default `functorch` implementations of many curvature quantities

wiseodd:
Changelog:
1. Add [Curvlinops](https://github.com/f-dangel/curvlinops/tree/main) backend.
2. Add default `functorch` implementation for many quantities: (in `CurvatureInterface` itself)
    1. Jacobians 
    2. Batch gradients
    3. Diag GGN (both stochastic and exact) and diag EF
    4. Full GGN (both stochastic and exact) and full EF

**Rationale for (1):** BackPACK & ASDL are not actively supported anymore and are hard to extend. Note that we use Curvlinops mainly for KFACs. 

**Rationale for (2):** `functorch` is a first-class PyTorch module => has the highest compatibility to any PyTorch models => futureproofing `laplace-torch`, less dependence to third-party backends

**Together:** They make `laplace-torch` very flexible and compatible with most architectures out there (and in the future).

With this PR, `Laplace(backend=CurvlinopsBackend)` supports:
1. Diag-GGN-exact, diag-GGN-MC, diag-EF for both classification and regression
2. Kron-GGN-exact, kron-GGN-MC, kron-EF for both classification and regression
3. Full-GGN-exact, full-GGN-MC, full-EF, full-Hessian-exact for both classification and regression

Note that previous backends only support some subsets of them.

I'd like to ask @runame to specifically review this since he's involved in Curvlinops. Specifically, for the K-FAC backend proposed here.

### DIFF
--- a/laplace/baselaplace.py
+++ b/laplace/baselaplace.py
@@ -4,10 +4,10 @@ import torch
 from torch.nn.utils import parameters_to_vector, vector_to_parameters
 from torch.distributions import MultivariateNormal, Dirichlet, Normal
 
-from laplace.utils import (parameters_per_layer, invsqrt_precision, 
+from laplace.utils import (parameters_per_layer, invsqrt_precision,
                            get_nll, validate, Kron, normal_samples,
                            fix_prior_prec_structure)
-from laplace.curvature import AsdlGGN, BackPackGGN, AsdlHessian
+from laplace.curvature import AsdlGGN, BackPackGGN, AsdlHessian, CurvlinopsGGN
 
 
 __all__ = ['BaseLaplace', 'ParametricLaplace',
@@ -62,7 +62,7 @@ class BaseLaplace:
         self.enable_backprop = enable_backprop
 
         if backend is None:
-            backend = AsdlGGN if likelihood == 'classification' else BackPackGGN
+            backend = CurvlinopsGGN
         self._backend = None
         self._backend_cls = backend
         self._backend_kwargs = dict() if backend_kwargs is None else backend_kwargs
@@ -357,7 +357,7 @@ class ParametricLaplace(BaseLaplace):
     """
 
     def __init__(self, model, likelihood, sigma_noise=1., prior_precision=1.,
-                 prior_mean=0., temperature=1., enable_backprop=False, 
+                 prior_mean=0., temperature=1., enable_backprop=False,
                  backend=None, backend_kwargs=None):
         super().__init__(model, likelihood, sigma_noise, prior_precision,
                          prior_mean, temperature, enable_backprop, backend, backend_kwargs)
@@ -527,7 +527,7 @@ class ParametricLaplace(BaseLaplace):
 
         return self.log_likelihood - 0.5 * (self.log_det_ratio + self.scatter)
 
-    def __call__(self, x, pred_type='glm', joint=False, link_approx='probit', 
+    def __call__(self, x, pred_type='glm', joint=False, link_approx='probit',
                  n_samples=100, diagonal_output=False, generator=None):
         """Compute the posterior predictive on input data `x`.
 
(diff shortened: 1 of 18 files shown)


laplace/curvature/curvlinops.py
```diff
@@ -0,0 +1,123 @@
...
+            separate_weight_and_bias=False
+        )
+        linop._compute_kfac()
+
+        kron = self._get_kron_factors(linop, len(y))
+        kron = self._rescale_kron_factors(kron, len(y), N)
+
+        loss = self.lossfunc(self.model(X), y)
+
+        return self.factor * loss.detach(), self.factor * kron
```
wiseodd, line 74: @runame please check this implementation. I'm not sure how to proceed. Comparing to BackPACK's & ASDL's KFAC, I still get the wrong results. 

If you need a test script: <https://gist.github.com/wiseodd/d9445133635a5d1d01f5b4c6e387586c>

laplace/curvature/curvlinops.py
```diff
@@ -0,0 +1,128 @@
...
+        kron *= self.factor
+
+        # Correction to match BackPACK & ASDL results
+        for factors in kron.kfacs:
+            for i in range(len(factors)):
+                factors[i] /= 2
+
+        loss = self.lossfunc(self.model(X), y)
+
+        return self.factor * loss.detach(), kron
```
wiseodd, line 79: @runame I found the culprit. Curvlinops' KFAC factors differ are twice as large as BackPACK's and ASDL's. So I put a correction in L72-75 above. 

I'm not sure what's happening. Please propose a more elegant fix.

runame (commented):
I reviewed the `curvlinops` part for now and pushed fixes directly.

I noticed that our tests are not systematic enough and that we are currently missing a few important cases (e.g., KFAC(-expand) is not the same with backpack and curvlinops for `Conv2d` modules). Since this branch still uses the old ASDL version, I think we should fix the tests of the curvature backends after this PR and #144 are merged.

laplace/curvature/curvlinops.py
runame: Fixed in [this commit](https://github.com/aleximmer/Laplace/pull/146/commits/8d8b8bb4eb90df05efcf79f6b8da8d67ecb29127). The issue was that `linop._compute_kfac()` was called twice, so that all Kronecker factors were added to itself once, resulting in the factor of 2. This can be avoided (also more efficient) by setting `check_deterministic=False`.

wiseodd:
@aleximmer please give a final check and feel free to merge!

laplace/baselaplace.py
```diff
@@ -791,7 +789,7 @@ def _init_H(self):
         self.H = torch.zeros(self.n_params, self.n_params, device=self._device)
 
     def _curv_closure(self, X, y, N):
-        return self.backend.full(X, y, N=N)
+        return self.backend.full(X, y)
```
aleximmer, line 792: Why does this have to be removed? I think there is reason for having it in case of subset-of-data approximations and it might be already used in Asdl. I would be in favor of not changing it unless necessary.

laplace/curvature/curvature.py
```diff
@@ -39,13 +40,18 @@ def __init__(self, model, likelihood, last_layer=False, subnetwork_indices=None)
...
+        self.params_dict = {k: v for k, v in self._model.named_parameters() if v.requires_grad}
+        self.buffers_dict = {k: v for k, v in self.model.named_buffers()}
 
     @property
     def _model(self):
         return self.model.last_layer if self.last_layer else self.model
 
     def jacobians(self, x, enable_backprop=False):
-        """Compute Jacobians \\(\\nabla_\\theta f(x;\\theta)\\) at current parameter \\(\\theta\\).
+        """
```
aleximmer, line 52: Comment should start right after `"""` as in all docstrings. What's the reason for the deviation here?

aleximmer (commented):
Looks overall good to me. Please have a look at the two comments and let me know what you think. After that it's good to go from my side.

Thanks for taking on the effort and including curvlinops!

laplace/curvature/curvature.py
```diff
@@ -188,38 +229,39 @@ class GGNInterface(CurvatureInterface):
...
-        loss : torch.Tensor
-        H_ggn : torch.Tensor
-            full GGN approximation `(parameters, parameters)`
-        """
-        loss = self.factor * self.lossfunc(f, y)
+            H_lik += 1/self.num_samples * torch.einsum('bc,bk->bck', grad_sample, grad_sample)
+
+        return H_lik
+
+    def _get_exact_H_lik(self, f):
```
aleximmer, line 257: Can we leave this is `_get_full_ggn` and add the function `_get_full_fisher` instead of these renamings? I `exact_H_lik` is inappropriate because it is not the true Hessian and would in principle have to be `log_lik`. `sample_H_lik` also sounds a bit different from what it actually is and just using `fisher` seems more appropriate to me.

aleximmer (commented):
Sorry, forgot to add this comment in the previous review.

laplace/curvature/curvature.py
wiseodd: I renamed them to better describe the functions. I think it's best to have them instead since there are many code overlaps in the `_get_full_ggn` and the proposed `_get_full_fisher` (also overlap with `diag_ggn` and `diag_fisher`).

Issue #45, opened 2021-10-20
Add methods to save and load Laplace instances


runame:
Users might want to avoid computing the Hessian approximation every time they run their code or reuse the same Laplace approximation in different files (e.g. #42). The best interface would probably be `.save(filepath)` and `.load(filepath)` methods.

wiseodd:
@runame @aleximmer do you have a sketch already on how to accomplish this? I'm willing to work on this. 

It will be very useful for LLMs etc since their forward passes are already quite expensive => fitting Laplace each time running a script is really tiring...

runame:
Addressed by #148.

PR #148, opened 2024-03-09
Add native serialization support

wiseodd:
Addressing #45. Very useful for large models like LLMs where even doing forward passes over training data (for `fit()`) is expensive.

The API basically follows PyTorch.

```python
la = Laplace(model, 'regression', subset_of_weights='all', hessian_structure='full')
la.fit(train_loader)
la.optimize_prior_precision()  # Or via marglik optimizing sigma_noise also

# Serialization for fitted quantities
state_dict = la.state_dict()
torch.save(state_dict, 'state_dict.bin')

la2 = Laplace(model, 'regression', subset_of_weights='all', hessian_structure='full')
# Load serialized, fitted quantities
la2.load_state_dict(torch.load('state_dict.bin'))
```

### DIFF
--- a/laplace/baselaplace.py
+++ b/laplace/baselaplace.py
@@ -989,6 +989,19 @@ class KronLaplace(ParametricLaplace):
         if len(self.prior_precision) not in [1, self.n_layers]:
             raise ValueError('Prior precision for Kron either scalar or per-layer.')
 
+    def state_dict(self) -> dict:
+        self._check_H_init()
+        state_dict = super().state_dict()
+        state_dict['H'] = self.H_facs.kfacs
+        return state_dict
+
+    def load_state_dict(self, state_dict: dict):
+        super().load_state_dict(state_dict)
+        self._init_H()
+        self.H_facs = self.H
+        self.H_facs.kfacs = state_dict['H']
+        self.H = self.H_facs.decompose(damping=self.damping)
+
 
 class LowRankLaplace(ParametricLaplace):
     """Laplace approximation with low-rank log likelihood Hessian (approximation).
--- a/laplace/baselaplace.py
+++ b/laplace/baselaplace.py
@@ -4,7 +4,7 @@ import torch
 from torch.nn.utils import parameters_to_vector, vector_to_parameters
 from torch.distributions import MultivariateNormal, Dirichlet, Normal
 
-from laplace.utils import (parameters_per_layer, invsqrt_precision, 
+from laplace.utils import (parameters_per_layer, invsqrt_precision,
                            get_nll, validate, Kron, normal_samples,
                            fix_prior_prec_structure)
 from laplace.curvature import AsdlGGN, BackPackGGN, AsdlHessian
@@ -357,7 +357,7 @@ class ParametricLaplace(BaseLaplace):
     """
 
     def __init__(self, model, likelihood, sigma_noise=1., prior_precision=1.,
-                 prior_mean=0., temperature=1., enable_backprop=False, 
+                 prior_mean=0., temperature=1., enable_backprop=False,
                  backend=None, backend_kwargs=None):
         super().__init__(model, likelihood, sigma_noise, prior_precision,
                          prior_mean, temperature, enable_backprop, backend, backend_kwargs)
@@ -527,7 +527,7 @@ class ParametricLaplace(BaseLaplace):
 
         return self.log_likelihood - 0.5 * (self.log_det_ratio + self.scatter)
 
-    def __call__(self, x, pred_type='glm', joint=False, link_approx='probit', 
+    def __call__(self, x, pred_type='glm', joint=False, link_approx='probit',
                  n_samples=100, diagonal_output=False, generator=None):
         """Compute the posterior predictive on input data `x`.
 
@@ -543,13 +543,13 @@ class ParametricLaplace(BaseLaplace):
 
         link_approx : {'mc', 'probit', 'bridge', 'bridge_norm'}
             how to approximate the classification link function for the `'glm'`.
-            For `pred_type='nn'`, only 'mc' is possible. 
+            For `pred_type='nn'`, only 'mc' is possible.
 
         joint : bool
             Whether to output a joint predictive distribution in regression with
             `pred_type='glm'`. If set to `True`, the predictive distribution
             has the same form as GP posterior, i.e. N([f(x1), ...,f(xm)], Cov[f(x1), ..., f(xm)]).
-            If `False`, then only outputs the marginal predictive distribution. 
+            If `False`, then only outputs the marginal predictive distribution.
             Only available for regression and GLM predictive.
 
         n_samples : int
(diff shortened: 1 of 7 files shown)


wiseodd:
By the way, I didn't really implement nor test `SubnetLaplace`. Not sure if it's really popular nowadays; the same effect can be done by setting `requires_grad = False` on select parameters anyway (from #144). 

Nevertheless, @edaxberger feel free to implement serialization on `SubnetLaplace`. It should be very straightforward for you.

wiseodd:
> By the way, I didn't really implement nor test `SubnetLaplace`. Not sure if it's really popular nowadays; the same effect can be done by setting `requires_grad = False` on select parameters anyway (from #144).
> 
> Nevertheless, @edaxberger feel free to implement serialization on `SubnetLaplace`. It should be very straightforward for you.

Though I must say that a quick test on `SubnetLaplace` works no problem. But I don't know about edge cases like different `subnetwork_indices` but the same `len(subnetwork_indices)`, etc.

laplace/lllaplace.py
```diff
@@ -103,12 +103,13 @@ def fit(self, train_loader, override=True):
         self.model.eval()
 
         if self.model.last_layer is None:
-            X, _ = next(iter(train_loader))
+            # Save an example batch for when loading the serialized Laplace
+            self.X, _ = next(iter(train_loader))
```
aleximmer, line 107: Could it make sense to save this rather as a shape and type instead of an actual sample that could be quite big? If it's too much effort, keeping it as is might be fine.

wiseodd: Good catch. I think it's better to do this since `X` is not alway a tensor. E.g. for HuggingFace models, this is `UserDict` with certain keys. For `likelihood='reward_modeling'`, this is also a `UserDict` but with other set of keys.

aleximmer: Ok so in this case one would have to apply something like a map over the userdict so it seems tricky. Seems fine to go this way then. I was rather thinking of other image or video problems where the input is on the order of the parameters. llms have very small inputs indeed.

wiseodd: Yeah, besides, `self.X` is on `cpu`. So, at least it won't incur burden on GPU. We can revisit this if specific use cases are requested by users.

laplace/baselaplace.py
```diff
@@ -768,6 +768,32 @@ def posterior_precision(self):
...
+        self._check_H_init()
+        state_dict = {
+            'mean': self.mean,
+            'H': self.H,
+            'loss': self.loss,
+            'prior_precision': self.prior_precision,
+            'sigma_noise': self.sigma_noise,
+            'n_data': self.n_data,
+            'n_outputs': self.n_outputs,
+        }
```
runame, line 781: I think there are a few things missing here, e.g. `model.state_dict()`, `likelihood`, `prior_mean`, `temperature`, `enable_backprop`, `backend`, `backend_kwargs`. Also, for the last layer classes `last_layer_name`. Finally, it would be good to store `subset_of_weights` and `hessian_structure` (see my main comment).

runame (changes requested):
Thanks for tackling this, will be very useful!

I think currently the assumption is that `model`, `likelihood`, `subset_of_weights`, and `hessian_structure` of `la` and `la2` (in the README example) match. It might make sense to explicitly check for this in the `load_state_dict` method and throw descriptive errors in case of a mismatch. The necessary information for this has to be also stored (see my comment below).

Additionally, it would be nice to allow for the creation of a new Laplace class instance *just based on the saved state dict*. A `classmethod` would make sense for this, but `Laplace` is a function and not a class, making this option impossible (without other changes). So alternatively, we could add a function like `LaplaceFromStateDict` that takes the state dict as an argument and returns a Laplace class instance which is (hopefully completely) equivalent to the previously saved one. I could add this functionality in a follow up PR though.

wiseodd:
@runame updated. Please check and merge if everything's good.

I decided to not include `backend` and `backend_kwargs` since Laplace doesn't care about them once `H` is obtained. I.e. `H` is just a tensor or `Kron` and you can use whatever backend you want for `glm_predictive`.

wiseodd:
I also ignore `model.state_dict` since the user would already have it anyway. Even marglik training also outputted the resulting torch model. Including it in the serialized Laplace is redundant.

runame (commented):
Thanks, will review the update soon!

I should have elaborated on this, but there is a very concrete use case where the `backend` and `backend_kwargs` are needed, i.e. whenever you want to set `fit(train_loader, override=False)`. For example, this is used to implement continual learning, see [our experiment](https://github.com/runame/laplace-redux/blob/main/continual_learning.py#L58) for the Redux paper. (I guess some mismatch between backends is fine, since only the shape (or both `Kron`) of `self.H` matters.)

I'm fine with not including `model.state_dict()` for now, this will only be important for the `LaplaceFromStateDict` functionality that I might add in a follow up PR.

laplace/baselaplace.py
```diff
@@ -961,6 +1020,18 @@ def prior_precision(self, prior_precision):
...
             raise ValueError('Prior precision for Kron either scalar or per-layer.')
 
+    def state_dict(self) -> dict:
+        state_dict = super().state_dict()
+        state_dict['H'] = self.H_facs.kfacs
+        return state_dict
+
+    def load_state_dict(self, state_dict: dict):
+        super().load_state_dict(state_dict)
+        self._init_H()
```
runame, line 1030: I think this will break the `override=False` use case. Maybe add a test for this case.

runame (commented):
Ok makes sense, only have one comment left.

wiseodd:
Added a test to check `fit(override=False)`!

tests/test_serialization.py
```diff
@@ -95,6 +95,25 @@ def test_serialize(laplace, model, reg_loader):
...
+    torch.save(la.state_dict(), 'state_dict.bin')
+
+    la2 = laplace(model, 'regression')
+    la2.load_state_dict(torch.load('state_dict.bin'))
+
+    # Emulating continual learning
+    la2.fit(reg_loader, override=False)
+
+    H_new = la2.H_kfacs.to_matrix() if laplace == KronLaplace else la2.H
+    assert not torch.allclose(H_orig, H_new)
```
runame, line 114: Could it even be `assert torch.allclose(2 * H_orig, H_new)`?

tests/test_serialization.py
```diff
@@ -95,6 +95,25 @@ def test_serialize(laplace, model, reg_loader):
     assert torch.allclose(f_var, f_var2)
 
 
+@pytest.mark.parametrize('laplace', set(flavors_no_llla) - {LowRankLaplace, KronLaplace, FullLaplace})
```
runame: Why is this is test only run for `DiagLaplace`?

runame (commented):
Sorry, last questions.

tests/test_serialization.py
wiseodd: It's an outdated commit. It's fixed in the latest one.

tests/test_serialization.py
wiseodd: Updated!

runame (approved):
Thanks for incorporating all feedback!

Issue #150, opened 2024-03-12
Failing tests: KFACLinearOperator object has no attribute _mapping


elcorto:
Hi

Thanks for this package. I have been testing the current `main` (fd4de8a8a2585230d1c276c32bd50307fa7e2276) with support for curvlinops. While `examples/regression_example.py` runs, my code and many (all?) `KronLaplace` tests fail with

```
laplace/baselaplace.py:907: in fit
    super().fit(train_loader, override=override)
laplace/baselaplace.py:410: in fit
    loss_batch, H_batch = self._curv_closure(X, y, N)
laplace/baselaplace.py:887: in _curv_closure
    return self.backend.kron(X, y, N=N)
laplace/curvature/curvlinops.py:72: in kron
    kron = self._get_kron_factors(linop)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <laplace.curvature.curvlinops.CurvlinopsGGN object at 0x2aab7f54f850>
linop = <122x122 KFACLinearOperator with dtype=float32>

    def _get_kron_factors(self, linop):
        kfacs = list()
        for name, module in self.model.named_modules():
>           if name not in linop._mapping.keys():
E           AttributeError: 'KFACLinearOperator' object has no attribute '_mapping'

laplace/curvature/curvlinops.py:39: AttributeError
```

(full log attached).

System info:

* asdfghjkl                         0.1a2
* backpack-for-pytorch              1.6.0
* curvlinops-for-pytorch            1.2.0
* torch                             2.1.1

[log.txt](https://github.com/aleximmer/Laplace/files/14571407/log.txt)

runame:
Hi Steve,

Can you clone the [`curvlinops`](https://github.com/f-dangel/curvlinops/) repo and install it from the `main` branch? This should resolve the issue.

We depend on the latest version of `curvlinops` that is not available on pypi yet. Sorry for the confusion, we are currently making major upgrades to our library and the `main` branch might be unstable for bit (but you'll get many new features).

elcorto:
> Hi Steve,
> 
> Can you clone the [`curvlinops`](https://github.com/f-dangel/curvlinops/) repo and install it from the `main` branch? This should resolve the issue.

That works, thanks.

[...]

PR #151, opened 2024-03-12
Point to curvlinops master branch in setup.cfg

wiseodd:
Useful when people do
```
pip install git+https://git@github.com/aleximmer/laplace
```
or 
```
git clone pip install https://github.com/aleximmer/laplace
cd laplace
pip install -e .
```

Addressing #150


runame (approved):
We can do this for now, but should switch back to the latest released version once we create a new `laplace-torch` release.

PR #159, opened 2024-03-24
Enable `torch.save()` + `torch.load()` for different `map_location`s

### LINKED ISSUE #158: Serialization: `torch.load(..., map_location="cpu")` doesn't map all tensors in `Laplace` instance
Hi

I came across an issue when serializing a `BaseLaplace`-derived object with tensors mapped to GPU and then reading it back in and mapping to CPU only. The same goes for a data structure containing such an object. See below for details on the use case.

When using `torch.load(..., map_location="cpu")` in

```py
import dill

lap = Laplace(
    model=model.to("cuda"),
    likelihood="regression",
    ...
)
lap.fit(train_dl)
torch.save(lap, "lap_gpu.pt", pickle_module=dill)

lap = torch.load("lap_gpu.pt", map_location="cpu")
```

one assumes that all tensors are mapped to CPU. This is true for weight tensors in `lap.model`, for instance. However, some tensor-valued properties in `Laplace`, such as `prior_precision_diag`, are calculated each time when accessed, and some of them access `self._device` for that. The latter is defined once in the `BaseLaplace` constructor

```py
self._device = next(model.parameters()).device
```

to be the same device as the one `model` lives on. `torch.load(..., map_location="cpu")` won't affect `self._device` and thus for all properties which use `self._device`, an attempt to calculate them will mix GPU and CPU tensors, which will fail.

One solution is to make `_device` a property.

```py
@property
def _device(self):
    return next(self.model.parameters()).device
```

I have a branch with this working. Before I open a PR, I'd like to know whether you think this is useful and if there are other implications that I did not encounter so far (test suite passes, of course). At least in terms of performance, this should be ok:

```py
>>> %timeit next(lap.model.parameters()).device
1.87 µs ± 2.73 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

Here are the combinations of `hessian_structure` and `subset_of_weights` that are affected (`load_ok=no`) or not (`load_ok=yes`).

```
hessian_structure subset_of_weights                                                             err load_ok
             diag        last_layer ValueError: Jacobians need to be on the same device as Laplace.      no
             kron        last_layer                                                                     yes
             full        last_layer                                                                     yes
             diag               all ValueError: Jacobians need to be on the same device as Laplace.      no
          lowrank               all                         RuntimeError: No HIP GPUs are available      no
             kron               all                                                                     yes
             full               all                         RuntimeError: No HIP GPUs are available      no
             full        subnetwork                         RuntimeError: No HIP GPUs are available      no
```

With the proposed fix, all of these work. I can provide more details and code to reproduce the above if needed.

The use case is serializing a fitted `Laplace` instance, including `self.model`, to disk and avoid the usual torch `load_state_dict` gymnastics (I'm aware of #148, but for small enough models and many test calculations, serializing stuff to one blob and loading it back in is more convenient). I may then want to run prediction later on other machines (in a cluster) by loading a fitted `Laplace` from disk and where CPU execution is fast enough. Disclaimer: I'm aware that serializing complex objects is not the best of practices because it has lots of potential pitfalls, and that this is not recommended for long-term data storage.

Still, I think it would be nice if `Laplace` instances behaved transparently when using `torch.load()` as `torch.Module`s do.

Thanks.

elcorto:
Closes #158

### DIFF
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -227,23 +227,23 @@ def test_serialize_fail_llla_different_last_layer_name(laplace, model, model3, r
 
 
 @pytest.mark.parametrize(
-    "model_device_str",
-    ["cpu", "cuda"] if torch.cuda.is_available() else ["cpu"],
+    'model_device_str',
+    ['cpu', 'cuda'] if torch.cuda.is_available() else ['cpu'],
 )
 @pytest.mark.parametrize(
-    "map_location",
-    ["cuda", "cpu"] if torch.cuda.is_available() else ["cpu"],
+    'map_location',
+    ['cuda', 'cpu'] if torch.cuda.is_available() else ['cpu'],
 )
-@pytest.mark.parametrize("laplace", flavors + flavors_subnet)
+@pytest.mark.parametrize('laplace', flavors + flavors_subnet)
 def test_map_location(
     laplace, model, reg_loader, tmp_path, map_location, model_device_str
 ):
     # Only for LLLaplace, we need dill.
     #   AttributeError: Can't pickle local object 'FeatureExtractor._get_hook.<locals>.hook'
     if issubclass(laplace, LLLaplace):
-        if find_spec("dill") is None:
+        if find_spec('dill') is None:
             pytest.skip(
-                reason="dill package not found but needed for this test"
+                reason='dill package not found but needed for this test'
             )
         else:
             import dill
@@ -257,7 +257,7 @@ def test_map_location(
         torch_save = lambda obj, fn: torch.save(obj, fn)
 
     device = torch.device(model_device_str)
-    kwds = dict(model=model.to(device), likelihood="regression")
+    kwds = dict(model=model.to(device), likelihood='regression')
     if issubclass(laplace, SubnetLaplace):
         kwds.update(subnetwork_indices=torch.LongTensor([1, 10, 104, 44]))
     la = laplace(**kwds)
@@ -270,7 +270,7 @@ def test_map_location(
     la.optimize_prior_precision()
     la.sigma_noise = 1231
 
-    save_fn = tmp_path / "la.pt"
+    save_fn = tmp_path / 'la.pt'
     torch_save(la, save_fn)
     la2 = torch.load(save_fn, map_location=map_location)
 
@@ -280,17 +280,17 @@ def test_map_location(
     # This is redundant if *all* will be used in the forward call below, where
     # things would fail if self._device is wrong.
     for name in [
-        "prior_precision_diag",
-        "prior_mean",
-        "prior_precision",
-        "sigma_noise",
+        'prior_precision_diag',
+        'prior_mean',
+        'prior_precision',
+        'sigma_noise',
     ]:
         assert (
             getattr(la, name).device.type == device.type
-        ), f"la.{name} failed"
+        ), f'la.{name} failed'
         assert (
             getattr(la2, name).device.type == map_location
-        ), f"la2.{name} failed"
+        ), f'la2.{name} failed'
 
     # Test tensor attrs.
     for name, obj in vars(la).items():
@@ -300,17 +300,17 @@ def test_map_location(
         # as in BaseLaplace.fit()). self.X won't be affected by self._device.
         # Since the predict() below passes, we ignore the wrong device here
         # for now.
-        if issubclass(laplace, LLLaplace) and name == "X":
+        if issubclass(laplace, LLLaplace) and name == 'X':
             continue
         if isinstance(obj, torch.Tensor):
-            assert obj.device.type == device.type, f"la.{name} failed"
+            assert obj.device.type == device.type, f'la.{name} failed'
             assert (
                 getattr(la2, name).device.type == map_location
-            ), f"la2.{name} failed"
+            ), f'la2.{name} failed'
 
     assert la.sigma_noise == la2.sigma_noise
     X, _ = next(iter(reg_loader))
     f_mean, f_var = la(X.to(device))
     f_mean2, f_var2 = la2(X.to(map_location))
-    assert torch.allclose(f_mean.to("cpu"), f_mean2.to("cpu"))
-    assert torch.allclose(f_var.to("cpu"), f_var2.to("cpu"))
+    assert torch.allclose(f_mean.to('cpu'), f_mean2.to('cpu'))
+    assert torch.allclose(f_var.to('cpu'), f_var2.to('cpu'))
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -3,6 +3,7 @@ import pytest
 from itertools import product
 import numpy as np
 from copy import deepcopy
+from importlib.util import find_spec
 import torch
 from torch import nn
 from torch.distributions.multivariate_normal import MultivariateNormal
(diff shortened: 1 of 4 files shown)


tests/test_serialization.py
```diff
@@ -217,3 +224,93 @@ def test_serialize_fail_llla_different_last_layer_name(laplace, model, model3, r
...
+    # Test tensor attrs.
+    for name, obj in vars(la).items():
+        # Only LLLaplace instances have a self.X that comes from the train
+        # loader, is assigned in fit() and not explicitly sent to device (it
+        # will be dynamically when used, as is done elsewhere in the code, such
+        # as in BaseLaplace.fit()). self.X won't be affected by self._device.
+        # Since the predict() below passes, we ignore the wrong device here
+        # for now.
+        if issubclass(laplace, LLLaplace) and name == "X":
+            continue
```
elcorto, line 304: We ignore the `self.X.device` check  on `LLLaplace` instances because of the reason described in the comment. Feedback from a core dev should clarify if this is OK or how else to handle this.

tests/test_serialization.py
```diff
@@ -217,3 +224,93 @@ def test_serialize_fail_llla_different_last_layer_name(laplace, model, model3, r
...
+        else:
+            import dill
+
+            torch_save = lambda obj, fn: torch.save(
+                obj, fn, pickle_module=dill
+            )
+    else:
+        # Use default pickle_module=pickle, but no need to import pickle here
+        # just to set pickle_module=pickle.
+        torch_save = lambda obj, fn: torch.save(obj, fn)
```
elcorto, line 257: The usage of `dill` for LLLaplace to pass tests is optional and those will be skipped with a message. Is this OK? I'd not make `dill` a requirement, but maybe the developers would like to see a warning here in addition to the skip?

elcorto (commented):
The PR is marked as WIP because of two questions I have in `test_serialization.py::test_map_location`. Thanks.

tests/test_serialization.py
```diff
@@ -217,3 +224,93 @@ def test_serialize_fail_llla_different_last_layer_name(laplace, model, model3, r
...
+
+    # device(type='cuda', index=0) != device(type='cuda')
+    ##assert la._device == device
+    assert la._device.type == device.type
+
+    la.fit(reg_loader)
+    la.optimize_prior_precision()
+    la.sigma_noise = 1231
+
+    save_fn = tmp_path / "la.pt"
```
wiseodd, line 273: Does this line create an actual file? If so, it should be cleaned up after this unit test finishes. You can add this file to the `cleanup` function above in this file.

tests/test_serialization.py
wiseodd: `self.X` is only being saved to infer the last-layer of the `self.model`. If I understand correctly, the `self.model` here is also serialized, right? In this case, we don't need to care about `self.X`, so I'm fine with skipping.

wiseodd (changes requested):
Thanks for the PR! I left some comments. Happy to merge after they're addressed.

tests/test_serialization.py
elcorto: Yes, `self.model` is serialized -- this is why `next(self.model.parameters()).device` returns the correct device. In the code, I only see `self.X` used [here](https://github.com/aleximmer/Laplace/blob/2543274aad34e2612d708a58c55272fadbc41626/laplace/lllaplace.py#L109-L113) and [here](https://github.com/aleximmer/Laplace/blob/2543274aad34e2612d708a58c55272fadbc41626/laplace/lllaplace.py#L181-L185) where it is sent to `self._device` if needed.

tests/test_serialization.py
elcorto: The supported API so far is `torch`-compatible as in

```py
torch.save(obj, filename)
```

So, `obj`, being a `Laplace` instance, would need to

* figure out that it is being serialized
* in that case, catch any `AttributeError` such as

  ```
  AttributeError: Can't pickle local object 'FeatureExtractor._get_hook.<locals>.hook'
  ```

  Here one must make the assumption that any `AttributeError` is actually caused by an attempted `torch.save()` call.

I understand the reasoning behind your suggestion. However I don't know, off the top of my head, how to implement this. I'm happy for any hints here.

One solution that comes to mind would be to have a `laplace.save()` function that wraps `torch.save()` but extends the API, kind of breaks `torch` compatibility and needs to be documented.

The other option is to document potential serialization fails with plain `pickle` and advise users to use `torch.save(..., pickle_module=dill)`. In either case, users must be made aware via documentation.

tests/test_serialization.py
```diff
@@ -217,3 +224,93 @@ def test_serialize_fail_llla_different_last_layer_name(laplace, model, model3, r
 
     with pytest.raises(ValueError):
         la2.load_state_dict(torch.load('state_dict.bin'))
+
+
+@pytest.mark.parametrize(
+    "model_device_str",
```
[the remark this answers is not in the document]
elcorto: Sure no problem. I did not find a coding style enforced (e.g. by `pre-commit`) or documented (I may have overlooked that), so I `black`-formatted my changes (`ruff format` actually). Please let me know where the coding rules are documented such that I can make changes accordingly.

tests/test_serialization.py
elcorto: It uses the `pytest` [`tmp_path` fixture])(https://docs.pytest.org/en/latest/reference/reference.html#tmp-path) so the file is subject to `pytest`'s file retention policy.

tests/test_serialization.py
wiseodd: Then let's go forward with ignoring `self.X.device` as you've already done.

tests/test_serialization.py
wiseodd: I'm leaning towards the documentation-based approach.

How about writing a new example specifically for serialization on README.md? Currently the Laplace-specific serialization example is coupled with another example:(https://github.com/aleximmer/Laplace/blob/2543274aad34e2612d708a58c55272fadbc41626/README.md?plain=1#L43-L76

Let's extract the serialization code from there, and add your `torch`-native-but-not-portable alternative. We can put this new example after  the `### Applying the LA over only a subset of the model parameters` example.

tests/test_serialization.py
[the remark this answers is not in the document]
wiseodd: We don't have a strict coding standard other than a single quote policy. IIRC we discussed this internally during the conception. 

Feel free to run formatting, but please run `black` with `--skip-string-normalization` when you do formatting.

tests/test_serialization.py
wiseodd: Ok good to know!

tests/test_serialization.py
elcorto: Agreed.

tests/test_serialization.py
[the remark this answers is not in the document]
elcorto: OK, will do, thanks.

tests/test_serialization.py
elcorto: I added a section to the README in 807644071078e8fb3bf41312c695dd1e762f4fa9 as you suggested. I feel that this may be a bit too technical for the README, should this instead go into the docs over at https://aleximmer.github.io/Laplace/?

wiseodd (approved):

wiseodd:
Thanks, @elcorto! Everything looks good. I'm fine with putting the explanation on the README---currently our docs website is admittedly not very readable.

I'm merging this --- thanks again for your contrib.

PR #168, opened 2024-04-23
Make the dict keys for models with dict-like inputs general

### LINKED ISSUE #165: Generalized dict-input mechanism
In #144, we added support for dict-like inputs. However, in `curvature/curvlinops.py` and `curvature/asdl.py`, the input tensor is assumed to be under `"input_ids"` key. We should improve upon it and make the key user specified (with default `"input_ids"`).

Note that these are the only place where an explicit `input_ids` key is assumed. It's used only to compute the batch size of the input tensor. So, should be straightforward to change.

wiseodd:
Closes #165 

@runame @aleximmer please wait until #144 is merged.


wiseodd:
Ready to review!

runame (approved):
LGTM

PR #172, opened 2024-04-25
Add an option to reduce LLM features in `LLLaplace`

### LINKED ISSUE #169: Handle feature reduction properly in LLLA
Say, you have an LLM with a regression head on top. Then this code

```python
f, phi = self.model.forward_with_features(x)
print(f.shape, phi.shape)
for p in self.model.last_layer.parameters():
    print(p.shape)
input()
```

outputs

```
torch.Size([4, 2]) torch.Size([4, 9, 768])
torch.Size([2, 768])
```

There is a mismatch between the last layer's inputs's dim and the last layer itself. 

The best solution seems to let the user pass what kind of reduction they use. Common choices: `first` (the `<CLS>` token in BERT), `last` (in causal LMs), `average` (https://arxiv.org/abs/2402.05015). We can use [enum](https://docs.python.org/3/library/enum.html) for this.

wiseodd:
Closes #169 

Wait for #144 to be merged.

I noticed that for Huggingface LLMs, the feature reduction is not automatically detected by `FeatureExtractor`'s hooks, unlike standard convnets. This is because HF actually reduces the logits themselves and not the features 😭 

In any case, this feature is useful as a "catch-all", making LLLA more general.


tests/test_lllaplace.py
```diff
@@ -27,6 +30,24 @@ def model():
...
+        def __init__(self):
+            super().__init__()
+            self.fc1 = nn.Linear(3, 20)
+            self.fc2 = nn.Linear(20, 2)
+            self.output_size = 2
+
+        def forward(self, x):
+            x = self.fc1(x)
+            x = nn.functional.relu(x)
+            x = x.mean(1)
```
runame, line 45: What's the reason this model only applies the mean reduction and doesn't have the option for `PICK_FIRST` or `PICK_LAST`?

laplace/utils/feature_extractor.py
```diff
@@ -66,6 +91,20 @@ def forward_with_features(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Te
         """
         out = self.forward(x)
         features = self._features[self._last_layer_name]
+
+        if len(features.shape) > 2 and self.feature_reduction is not None:
```
runame, line 95: Nit: `len(features.shape)` --> `features.dim()`

laplace/utils/feature_extractor.py
```diff
@@ -66,6 +91,20 @@ def forward_with_features(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Te
...
+            n_intermediate_dims = len(features.shape) - 2
+
+            if self.feature_reduction == FeatureReduction.PICK_FIRST:
+                features = features[:, *([0] * n_intermediate_dims), :].squeeze()
+            elif self.feature_reduction == FeatureReduction.PICK_LAST:
+                features = features[:, *([0] * n_intermediate_dims), :].squeeze()
+            else:
```
runame, line 102: Maybe check explicitly `if self.feature_reduction == FeatureReduction.AVERAGE` and add an else-statement that raises a `ValueError`?

runame (approved):
LGTM

tests/test_lllaplace.py
wiseodd: The options in `Laplace` are completely detached to the one made by the model (`Laplace` has no way knowing that). The user has to take care of this themselves. It is noted both here
https://github.com/aleximmer/Laplace/blob/ff139fe936148fc5e7ace56b26917b453d75e0bc/laplace/lllaplace.py#L61-L63
and here
https://github.com/aleximmer/Laplace/blob/ff139fe936148fc5e7ace56b26917b453d75e0bc/examples/huggingface_example.py#L106-L108


But in any case, for test purposes, setting the mock model to have just `mean` is sufficient.

PR #177, opened 2024-04-26
Replace hardcoded class index with `logit_class_dim` argument

### LINKED ISSUE #163: Conform to PyTorch convention in the loss
Currently we always assume that the class index in the outputs of the model to be `-1`. While this works for standard models and Huggingface LLMs, there are some important models where this is false. E.g. image-output models where the logit tensor is of shape `(n_batch, n_classes, height, width)`, i.e. the class index is `1`.

wiseodd:
Closes #163 

Please wait until #144 is merged.


wiseodd:
Still WIP test cases

wiseodd:
Ready to review!

Discussion points:
1. Do BackPACK, ASDL, Asdfghjkl even support multiple output dims? That is, if we flatten `logits = logits.view(-1, logits.size(logit_class_dim))`, do they even compute the correct quantities? 
2. Curvlinops and torch.func interfaces always assume `logit_class_dim = 1`. Do we want to make them respect `logit_class_dim`? I don't think flattening as above is the correct approach, right?

wiseodd:
Merged with `main` and ready for review!

tests/test_curv_backends_interface.py
```diff
@@ -361,3 +408,92 @@ def test_full_ef_reg_curvlinops_against_backpack_full(reg_Xy, model):
...
+        AsdfghjklGGN,
+        AsdfghjklEF,
+        AsdfghjklHessian,
+        BackPackGGN,
+        BackPackEF,
+    ],
+)
+@pytest.mark.parametrize("method", ["full", "kron", "diag"])
+@pytest.mark.parametrize("logit_class_dim", [-1, 1000])
+def test_logit_class_dim_class(backend_cls, method, logit_class_dim, model, class_Xy):
```
runame, line 431: I think this has to be tested for actual equivalence of the computation, i.e. `pytest.mark.parametrize` the tests of the backends with `logit_class_dim`.

runame (changes requested):
1. I think @f-dangel mentioned that BackPack doesn't support multiple output dims, so we have to rearrange to the equivalent 2d output/label. ASDL assumes the class dim is `-1` and flattens everything, so again, we have to rearrange to the equivalent 2d output. Same probably holds for asdfghjkl. With other words, we can't just flatten the output, but have to transpose correctly (probably most readable with `einops.rearrange`). See [here](https://github.com/f-dangel/curvlinops/blob/main/curvlinops/kfac.py#L610-L616), for how we handle this in curvlinops with assumption that `logit_class_dim=1` for CE loss and `logit_class_dim=-1` for MSE loss.
2. For functorch, it should be possible to also rearrange the logits/labels. For curvlinops, the issue is that we don't have access to the logits, so to make it work here we would have to wrap the forward pass of the model to fix the logit class dim at the end of the forward pass. So maybe we should just raise an informative error, telling the user how to modify the forward pass of the model to be compatible with curvlinops.

To make sure we get this right, we have to test for equivalence of computation, see my comment. Also, if we only want `logit_class_dim` to affect classification tasks, we should maybe verify at initialization that `likelihood="classification"` if `logit_class_dim` is not the default value and raise a `ValueError` otherwise.

wiseodd:
It seems more complicated than anticipated. This PR is useful for models with image outputs like diffusion models. 

Considering v0.2 is all about LLMs, let's defer this to v0.3!

runame:
> It seems more complicated than anticipated. This PR is useful for models with image outputs like diffusion models.
> 
> Considering v0.2 is all about LLMs, let's defer this to v0.3!

Ok, for now maybe we can add a note in the README and the docstring that clearly states how multi dim outputs are handled?

wiseodd:
So what do you have in mind regarding the wording? Something like this in README.md?

```markdown
## Caveats

- Currently, this library always assumes that the model has an 
  output tensor of shape `(batch_size, ..., n_classes)`, so in 
  the case of image outputs, you need to rearrange from NCHW to NHWC.
```

runame:
> So what do you have in mind regarding the wording? Something like this in README.md?

Yes, this is exactly what I was thinking!

PR #182, opened 2024-04-27
Make `CurvatureInterface.functorch_jacobians` simply an alias to `CurvatureInterface.jacobians`

wiseodd:
Before, they were duplicates. The rationale is to make `functorch_jacobians` accessible even when the subclasses overloaded `jacobians`. Useful for Bayesian optimization since not all backend's `jacobians` are backpropable (e.g. ASDL detaches everything).


runame (commented):
Not sure how this relates to #187, was that PR based on this, i.e. supersedes it?

wiseodd:
Let's forget about this due to #187. The next plan is to decouple the Hessian (for fit) and Jacobian (for GLM) computations.

PR #183, opened 2024-04-29
Move back to ASDL's main repo as dependency

aleximmer:
Kazuki merged the changes by @wiseodd to asdl so we can move back to the original repo as dep.


wiseodd (approved):
LGTM

PR #180, opened 2024-04-27
Typehinting

wiseodd:
Closes #68 

WIP and blocked by #144

PR #176, opened 2024-04-26
Add `ruff check`, `ruff format --check`, and `pytest` Github actions

wiseodd:
Closes #152 

Self-explanatory!

PR #167, opened 2024-04-23
Doc & example for reward modeling

wiseodd:
Closes #166 

Self explanatory. @aleximmer @runame

PR #192, opened 2024-06-04
Functional Laplace Updated

Ludvins:
Took @metodmove's contribution from [PR 55](https://github.com/aleximmer/Laplace/pull/55) and adapted it to the current state of the repository. 

Functional Laplace can be used with `hessian='gp'`, there is an example at [calibration_gp_example.py](https://github.com/Ludvins/Laplace/blob/main/examples/calibration_gp_example.py) and [calibration_gp_example.md](https://github.com/Ludvins/Laplace/blob/main/examples/calibration_gp_example.md).

Two unit-tests are available: [test_functional_laplace.py](https://github.com/Ludvins/Laplace/blob/main/tests/test_functional_laplace.py) and [test_functional_laplace_unit.py](https://github.com/Ludvins/Laplace/blob/main/tests/test_functional_laplace_unit.py).


Ludvins:
> Thanks a lot for the contribution! Looks great at a glance!
> 
> I will review it in-depth this weekend. But for now, could you please make sure it conforms to both `ruff` linter and autoformatter? We plan to be strict about this in v0.2, see PR #176, so might as well start enforcing it now :) (Additionally, `isort` autoformat wouldn't hurt.)
> 
> @aleximmer should also review this since this is based on his papers.

I'll handle the formatting, should have done it before the PR. Do you want me to do it to the whole project or just the files regarding the PR?

wiseodd:
No worries! We should've been explicit about contributing in the README in the first place.

The linting and autoformatting I requested are just for the scope of this PR. The rest has already been done in PR #176.

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
 
 
 class FunctionalLaplace(BaseLaplace):
-    pass
+    """
```
wiseodd, line 1570: There shouldn't be `\n` after `"""`: https://peps.python.org/pep-0257/#multi-line-docstrings

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+    resort to such an approximation because of the (possible) errors found in Laplace approximation for
+    multiclass GP classification in Chapter 3.5 of [R&W 2006 GP book](http://www.gaussianprocess.org/gpml/),
+    see the question
+    [here](https://stats.stackexchange.com/questions/555183/gaussian-processes-multi-class-laplace-approximation)
+    for more details. Alternatively, one could also resort to *one-vs-one* or *one-vs-rest* implementations
+    for multiclass classification, however, that is not (yet) supported here.
+
+    Parameters
+    ----------
+    M : int
```
wiseodd, line 1593: I prefer a more descriptive name for this parameter. Maybe `num_data`?

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+        sigma_noise=1.0,
+        prior_precision=1.0,
+        prior_mean=0.0,
+        temperature=1.0,
+        backend=BackPackGGN,
+        backend_kwargs=None,
+        enable_backprop=False,
+        diagonal_kernel=False,
+        seed=0,
+    ):
```
wiseodd, line 1621: Could you please add typehinting for all parameters here and everywhere else? We're in the process of enforcing this for the next release https://github.com/aleximmer/Laplace/pull/180.

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+        assert backend in [BackPackGGN, AsdlGGN]
```
wiseodd, line 1622: I wonder if `CurvlinopsGGN` also works here

laplace/baselaplace.py
wiseodd: (It's fine to use `M` internally, but for a user-facing interface, a descriptive name is better)

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+        self.train_loader = (
+            None  # needed in functional variance and marginal log likelihood
+        )
+        self.batch_size = None
+        self._prior_factor_sod = None
+        self.mu = None  # mean in the scatter term of the log marginal likelihood
+        self.L = None
+
+        # MAP estimate of NN parameters (used in regression marginal likelihood)
+        self.map_estimate = parameters_to_vector(self.model.parameters()).detach()
```
wiseodd, line 1651: I believe `BaseLaplace` already store the same thing in `self.mean`. If possible, avoid duplicates to keep memory requirements low

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+                device=self._device,
+            )
+
+    def _store_K_batch(self, K_batch, i, j):
+        """
+        Given the kernel matrix between the i-th and the j-th batch, stores it in the
+        corresponding position in self.K_MM.
+        """
+        if self.diagonal_kernel:
+            for c in range(self.n_outputs):
```
wiseodd, line 1713: Is it possible to avoid the loop?

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+            shuffle=False,
+        )
+
+    def fit(self, train_loader, progress_bar=False):
+        """Fit the Laplace approximation of a GP posterior.
+
+        Parameters
+        ----------
+        train_loader : torch.data.utils.DataLoader
+            `train_loader.dataset` needs to be set to access \\(N\\), size of the data set
```
wiseodd, line 1808: Missing docstring for `progress_bar`

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+            self.likelihood == 'regression'
+            and self.n_outputs > 1
+            and self.diagonal_kernel
+        ):
+            warnings.warn(
+                'Using FunctionalLaplace with the diagonal approximation of a GP kernel is not recommended '
+                'in the case of multivariate regression. Predictive variance will likely be overestimated.'
+            )
+
+        self.model.eval()
```
wiseodd, line 1828: Does this mean the operations above needs `self.model` to be in the training mode?

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+
+        N = len(train_loader.dataset)
+        self.n_data = N
+
+        # by default, all training data points are used for GP inference
```
wiseodd, line 1833: I wonder if this is a good default, in terms of user perception, since this will be very slow. Does it make sense to make `M` mandatory (i.e. not allowing for `None`---the user must specify it)?

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+        self._init_Sigma_inv()
+
+        f, lambdas, mu = [], [], []
+
+        if progress_bar:
+            loader = enumerate(tqdm.tqdm(train_loader, desc='Fitting'))
+        else:
+            loader = enumerate(train_loader)
+
+        for i, (X, y) in loader:
```
wiseodd, line 1851: Do you see this `FunctionalLaplace` applicable (read: fast enough, thus useful) to foundation models like LLMs? If so, might consider allowing for custom `MutableMapping` for `X`.

See for example: https://github.com/aleximmer/Laplace/blob/main/examples/huggingface_example.md
Another example for preference learning / reward modeling: https://github.com/aleximmer/Laplace/pull/167

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+            _, K = f_mu.size(0), f_mu.size(-1)
+            f_var_diag = torch.diagonal(f_var, dim1=1, dim2=2)
+            # optional: variance correction
+            if link_approx == 'bridge_norm':
+                f_var_diag_mean = f_var_diag.mean(dim=1)
+                f_var_diag_mean /= torch.as_tensor([K / 2], device=self._device).sqrt()
+                f_mu /= f_var_diag_mean.sqrt().unsqueeze(-1)
+                f_var_diag /= f_var_diag_mean.unsqueeze(-1)
+            sum_exp = torch.exp(-f_mu).sum(dim=1).unsqueeze(-1)
+            alpha = (1 - 2 / K + f_mu.exp() / K**2 * sum_exp) / f_var_diag
```
wiseodd, line 2022: Let's extract these into some util files. They are now duplicates of the ones in `ParametricLaplace`.

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+              NN output (batch, output_shape)
+        y: torch.tensor
+              data labels (batch, output_shape)
+
+        Returns
+        -------
+        mu : torch.tensor
+            K_batch_star with shape (batch, output_shape)
+        """
+
```
wiseodd, line 2438: No empty line after docstring

laplace/lllaplace.py
```diff
@@ -308,3 +320,95 @@ class DiagLLLaplace(LLLaplace, DiagLaplace):
...
+class FunctionalLLLaplace(FunctionalLaplace):
+    """Here not much changes in terms of GP inference compared to FunctionalLaplace class.
+    Since now we treat only the last layer probabilistically and the rest of the network is used as a "fixed feature
+    extractor", that means that the \\(X \in \mathbb{R}^{M \\times D}\\) in GP inference changes
+    to \\(\\tilde{X} \\in \mathbb{R}^{M \\times l_{n-1}} \\),  where \\(l_{n-1}\\) is the dimension of the output
+    of the penultimate NN layer.
+
+    See `FunctionalLaplace` for the full interface.
+    """
+
```
wiseodd, line 334: No empty line after docstring

laplace/lllaplace.py
```diff
@@ -308,3 +320,95 @@ class DiagLLLaplace(LLLaplace, DiagLaplace):
...
+        """
+        self.model.eval()
+
+        if self.model.last_layer is None:
+            X, _ = next(iter(train_loader))
+            with torch.no_grad():
+                self.model.find_last_layer(X[:1].to(self._device))
+            self.map_estimate = parameters_to_vector(
+                self.model.last_layer.parameters()
+            ).detach()
```
wiseodd, line 401: Check if this already present in the form of `self.mean` or similar

laplace/utils/utils.py
```diff
@@ -219,6 +219,19 @@ def block_diag(blocks):
     return M
 
 
+class SoDSampler(Sampler):
+
+    def __init__(self, N, M, seed: int = 0):
+        np.random.seed(seed)
+        self.indices = torch.tensor(np.random.choice(list(range(N)), M, replace=False))
```
wiseodd, line 226: Need a docstring

wiseodd, line 226: Does it make sense to use pure PyTorch here? In this case, `torch.Generator` use be used instead. Globally changing the random seed might have side effect at places far away

tests/test_functional_laplace.py
```diff
@@ -0,0 +1,151 @@
...
+    toy_classification_dataset,
+    toy_model,
+    toy_multivariate_regression_dataset,
+    toy_regression_dataset_1d,
+)
+
+true_sigma_noise = 0.1
+torch.manual_seed(711)
+
+# torch.set_default_dtype(torch.float64)
```
wiseodd, line 25: Remove comment

tests/test_functional_laplace.py
```diff
@@ -0,0 +1,151 @@
...
+    full_la.fit(train_loader)
+    functional_gp_la.fit(train_loader)
+
+    p_full = full_la(X_test)
+    p_gp = functional_gp_la(X_test)
+
+    p_full = p_full.squeeze().detach().cpu().numpy()
+    p_gp = p_gp.squeeze().detach().cpu().numpy()
+
+    # difference due to the diagonal approximation of L
```
wiseodd, line 144: Remove debugging comments

tests/test_functional_laplace.py
```diff
@@ -0,0 +1,151 @@
...
+    # diffs = np.abs(p_full - p_gp)
+    # print(diffs.mean())
+    # print(diffs.max())
+    # print(f"{(np.argmax(p_full, axis=1) == np.argmax(p_gp, axis=1)).sum()} out of {p_full.shape[0]}")
+
+    assert p_full.shape == p_gp.shape
+    assert np.array_equal(np.argmax(p_full, axis=1), np.argmax(p_gp, axis=1))
```
wiseodd, line 151: Is `torch.allclose` not working here? Or is there another reason for using numpy?

laplace/baselaplace.py
```diff
@@ -1568,8 +1567,1025 @@ def sample(self, n_samples=100, generator=None):
...
+                )
+        else:
+            raise ValueError(
+                'Prior precision either scalar or torch.Tensor up to 1-dim.'
+            )
+        # This is a change from BaseLaplace. If the prior precision is changed, the cholesky
+        #  decomposition needs to be recomputed.
+        self._recompute_Sigma = True
+
+    def state_dict(self) -> dict:
```
wiseodd, line 2508: Please add a test case for saving & loading `FunctionalLaplace` in `tests/test_serialization.py`

wiseodd (changes requested):
Thanks again @Ludvins for the great PR! 

@aleximmer please review the method implementation in more detail.

laplace/baselaplace.py
Ludvins: Actually yes! Changed that!

laplace/baselaplace.py
Ludvins: No, I moved it to the first instruction in the function.

laplace/baselaplace.py
Ludvins: I do not have a strong opinion here. I'll change it to mandatory.

laplace/baselaplace.py
Ludvins: I noticed that in the ParametricLaplace `__call__` method with MC link, `_glm_predictive_distribution` is called twice: one in `__call__` and another time in `predictive_samples`. In FunctionalLaplace `predictive_samples` is taking directly `f_mu` and `f_var` to avoid this. However, that makes the function difficult to use directly. Should I re-factor the `__call__` function in ParametricLaplace to avoid the double computation and move it to an utils file?

laplace/utils/utils.py
Ludvins: I believe the only way to do this with torch is using a permutation as `self.indices = torch.randperm(N, generator=rng)[:M]`

tests/test_functional_laplace.py
Ludvins: I think the original tests were using numpy instead of torch. I can change it if you prefer to use `torch.allclose`.

examples/helper/models/FMNIST_CNN_10_2.2e+02.pt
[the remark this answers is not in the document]
Ludvins: I agree, is there anywhere I can upload the file? It was already in metod's PR branch when I took it.

laplace/baselaplace.py
Ludvins: I believe these loops where added by Metod to solve (or mitigate) memory problems in the GP approach. See [commit in PR](https://github.com/metodj/Laplace/pull/2/commits/ac74a0b8b6517b297d73ea12e4285d33461a99f5). I could revert them but those problems would (probably) raise again.

laplace/baselaplace.py
Ludvins: Probably not useful in practice but I have adapted the code to allow it

laplace/baselaplace.py
wiseodd: Ok, makes sense 👌

laplace/baselaplace.py
```diff
@@ -1781,8 +1786,1062 @@ def sample(
...
+              NN output (batch, output_shape)
+        y: torch.tensor
+              data labels (batch, output_shape)
+
+        Returns
+        -------
+        mu : torch.tensor
+            K_batch_star with shape (batch, output_shape)
+        """
+
```
wiseodd, line 2695: Not sure if a problem with Github's diff interface. But there is still an empty line here between the docstring and the function body. If this is true, please remove it.

laplace/curvature/backpack.py
```diff
@@ -94,9 +94,14 @@ def jacobians(
         CTX.remove_hooks()
         _cleanup(model)
         if model.output_size > 1:
-            return torch.stack(to_stack, dim=2).transpose(1, 2), f
+            J = torch.stack(to_stack, dim=2).transpose(1, 2)
         else:
-            return Jk.unsqueeze(-1).transpose(1, 2), f
+            J = Jk.unsqueeze(-1).transpose(1, 2)
+
+        if not enable_backprop:
```
wiseodd, line 101: Fits a single line:

```diff
-if not enable_backprop:
-     f = f.detach()
-     J = J.detach()
-return J, f

+return (J, f) if enable_backprop else (J.detach(), f.detach())
```

laplace/utils/utils.py
wiseodd: I see. The main issue for me is that it changes the global rng state. Can you use Numpy's generator?

https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.choice.html#numpy.random.Generator.choice

tests/test_functional_laplace.py
wiseodd: Yes let's use `torch.allclose` instead and remove the `.numpy()`'s above.

wiseodd (changes requested):

laplace/baselaplace.py
wiseodd: Yes, would be great if you could refactor them.

wiseodd:
Thanks for the changes! The code looks good now and passes all checks. Only small bits of changes needed this time.

wiseodd:
@aleximmer please review if you have chance.

Ludvins:
I've created two functions `_glm_forward_call` and `_glm_predictive_samples` in BaseLaplace that:
1. Avoid the double computation of the glm predictive that was present in ParametricLaplace when using predictive samples.
2. Is used for FunctionalLaplace and avoids repeating too much code in Parametric and Functional Laplace.

This can be moved as an utils function but it might be usefull in BaseLaplace too.

wiseodd (approved):
LGTM. Once @aleximmer gives his reviews we can merge

laplace/baselaplace.py
```diff
@@ -1792,8 +1881,1041 @@ def sample(
...
+            temperature,
+            enable_backprop,
+            dict_key_x,
+            dict_key_y,
+            backend,
+            backend_kwargs,
+        )
+        self.enable_backprop = enable_backprop
+
+        self.num_data = num_data
```
aleximmer, line 1953: Would call this `n_data_sod`, `n_data_subset`, or `n_subset`. Later, you compare `n_data` to `num_data` and that is confusing since both refer potentially to the same thing.

laplace/baselaplace.py
```diff
@@ -1792,8 +1881,1041 @@ def sample(
...
+    see the question
+    [here](https://stats.stackexchange.com/questions/555183/gaussian-processes-multi-class-laplace-approximation)
+    for more details. Alternatively, one could also resort to *one-vs-one* or *one-vs-rest* implementations
+    for multiclass classification, however, that is not (yet) supported here.
+
+    Parameters
+    ----------
+    num_data : int
+        number of data points for Subset-of-Data (SOD) approximate GP inference.
+    diagonal_kernel : bool
```
aleximmer, line 1908: I feel like `independent_kernel` or `independent_outputs` would be more suitable variable name. `diagonal_kernel` sounded to me like the extreme case a subset of a single data point so no correlations are estimated. What do you think?

aleximmer (commented):
Thanks for the great effort. Please see the comments regarding the two variable names and let's discuss them. Other than that, it's ready to merge.

laplace/baselaplace.py
Ludvins: Totally agree, I think `n_subset` is a good name for the variable.

laplace/baselaplace.py
Ludvins: It is true the name might be confusing, I prefer `independent_outputs` rather than `independent_kernel`  as in the latter it might be unclear what "independent" means.

laplace/baselaplace.py
```diff
@@ -1821,8 +1909,1041 @@ def sample(
...
+            is returned with the mean and the predictive covariance.
+        """
+        if self._fitted is False:
+            raise RuntimeError(
+                "Functional Laplace has not been fitted to any "
+                + "training dataset. Please call .fit method."
+            )
+
+        if self._recompute_Sigma is True:
+            Warning(
```
aleximmer, line 2336: Before, you use `warnings.warn` and here just `Warning`. What is the difference? Please make it consistent. According to ruff, it should be `raise Warning` by the way.

laplace/baselaplace.py
```diff
@@ -1821,8 +1909,1041 @@ def sample(
...
+        self.n_outputs = out.shape[-1]
+        setattr(self.model, "output_size", self.n_outputs)
+        self.batch_size = train_loader.batch_size
+
+        if (
+            self.likelihood == "regression"
+            and self.n_outputs > 1
+            and self.independent_outputs
+        ):
+            warnings.warn(
```
aleximmer, line 2188: This is the "before" I mean :)

Issue #194, opened 2024-06-11
RunningNLLMetric returns infinite when running optimize_prior_precision


MJordahn:
Discussion started in #193.

After running `fit` (using the `AsdlGGN ` backend) in `FullLLLaplace`, I can do predictions using the Laplace module (and see that my predictive function changes in comparison to my MAP model - but in a reasonable way). When I then run `optimize_prior_precision` using the `gridsearch` method, the models predictive function breaks.

When I use the `progress_bar` option in the Github version of the repo I get the following (just a snippet):

```
  0%|          | 0/21 [00:00<?, ?it/s]
[Grid search | prior_prec: 1.000e-04, loss: inf]:   0%|          | 0/21 [00:22<?, ?it/s]
[Grid search | prior_prec: 1.000e-04, loss: inf]:   5%|▍         | 1/21 [00:22<07:39, 22.97s/it]
[Grid search | prior_prec: 2.512e-04, loss: inf]:   5%|▍         | 1/21 [00:44<07:39, 22.97s/it]
[Grid search | prior_prec: 2.512e-04, loss: inf]:  10%|▉         | 2/21 [00:44<07:00, 22.12s/it]
[Grid search | prior_prec: 6.310e-04, loss: inf]:  10%|▉         | 2/21 [01:07<07:00, 22.12s/it]
[Grid search | prior_prec: 6.310e-04, loss: inf]:  14%|█▍        | 3/21 [01:07<06:43, 22.43s/it]
[Grid search | prior_prec: 1.585e-03, loss: inf]:  14%|█▍        | 3/21 [01:30<06:43, 22.43s/it]
[Grid search | prior_prec: 1.585e-03, loss: inf]:  19%|█▉        | 4/21 [01:30<06:23, 22.57s/it]
[Grid search | prior_prec: 3.981e-03, loss: inf]:  19%|█▉        | 4/21 [01:52<06:23, 22.57s/it]
[Grid search | prior_prec: 3.981e-03, loss: inf]:  24%|██▍       | 5/21 [01:52<06:02, 22.64s/it]
[Grid search | prior_prec: 1.000e-02, loss: inf]:  24%|██▍       | 5/21 [02:15<06:02, 22.64s/it]
[Grid search | prior_prec: 1.000e-02, loss: inf]:  29%|██▊       | 6/21 [02:15<05:40, 22.70s/it]
[...]

PR #199, opened 2024-06-14
Remove `try-except` from gridsearch 

### LINKED ISSUE #157: Make try-except statement in `_gridsearch` more specific
Currently, if there is, e.g., any `RuntimeError` in the `loss`, `_gridsearch` will silently fail and just return the boundary value of the grid; see [here](https://github.com/aleximmer/Laplace/blob/main/laplace/baselaplace.py#L288-L302). One example for this is when the loss function is not appropriate for the setting.

wiseodd:
Fixes #157 and #194 

I added new test cases and also made sure to use the classification likelihood during gridsearch in the case of reward modeling.


laplace/baselaplace.py
```diff
@@ -429,21 +438,20 @@ def _gridsearch(
         results = list()
         prior_precs = list()
         pbar = tqdm.tqdm(interval) if progress_bar else interval
+
         for prior_prec in pbar:
             self.prior_precision = prior_prec
-            try:
```
runame, line 434: Maybe it's worth keeping the try-except statement for its original purpose, to catch inversion errors for a prior precision that is too small, so that the gridsearch is not interrupted, but the next value is tried. So it could be something like this (this would work when using Cholesky for the inversions):
```python
try:
    ...
except RuntimeError as error:
    if "not positive-definite" in str(error):
        result = np.inf
    else:
        raise error
```

runame (commented):
See #200 for a suggested refactor and my comment below.

laplace/baselaplace.py
```diff
@@ -442,8 +454,13 @@ def _gridsearch(
                     loss_with_var=loss_with_var,
                     dict_key_y=self.dict_key_y,
                 )
-            except RuntimeError:
+            except LinAlgError:
```
wiseodd, line 457: @runame please recheck. It seems PyTorch now uses `LinAlgError` for this https://github.com/pytorch/pytorch/pull/68571. But I also added your suggestion for safety.

runame (approved):
LGTM now!

PR #200, opened 2024-06-15
Refactor reward-modeling likelihood

runame:
Suggested refactor of the reward-modeling likelihood, let me know if I missed something. Advantages:

1. Simplicity and readability. We get rid of `self.reward_modeling` and `self._fitting` (actually `self._fitting` was not even used before, not sure if you had a future use case in mind?).
2. When a user calls `la.likelihood` it will always return `"reward_modeling"` which should be the expected behavior.

(Unrelated: I also removed a superfluous `loss_with_var` argument.)

### DIFF
--- a/laplace/baselaplace.py
+++ b/laplace/baselaplace.py
@@ -148,6 +148,11 @@ class BaseLaplace:
         self.n_outputs = None
         self.n_data = 0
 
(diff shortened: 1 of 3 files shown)


wiseodd (approved):
LGTM. Thanks for the refactor, definitely makes things cleaner now!

PR #202, opened 2024-07-04
Caveats

### LINKED ISSUE #147: Fix defaults and suggestions
In contrast to the redux paper, we give an example with post-hoc Laplace and marginal likelihood prior precision tuning in the docs (https://aleximmer.github.io/Laplace/#full-example-post-hoc-laplace-on-a-large-image-classifier). This is confusing and we have not thoroughly tested this approach but only recommended full + online marglik or last-layer + validation-based prior tuning.

### LINKED ISSUE #193: Fit function of FullLLLaplace runs slower in Github version in comparison to pip version
Hello,

After having installed the new version of Laplace as specified in [#191 ](https://github.com/aleximmer/Laplace/issues/191#issuecomment-2147695295), I am getting much slower fitting times using the `FullLLLaplace` module. I use the exact same code base only exchanging the version of Laplace (and new dependencies of the Github version). Running on the same GPU (Titan V) I time the `fit` function to take approximately 10 seconds in the pip version and 1805 seconds in the Git. I am also getting a user warning in the Git version, which I do not get using the pip version:

```
/home/mikkjo/.conda/envs/ude/lib/python3.12/site-packages/torch/autograd/graph.py:744: UserWarning: Attempting to run cuBLAS, but there was no current CUDA context! Attempting to set the primary context... (Triggered internally at ../aten/src/ATen/cuda/CublasHandlePool.cpp:135.)
  return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
```

Am I supposed to be calling the `init` or `fit` functions differently in the Git version which may be the cause of the difference in fit time?

wiseodd:
Closes #147 
Closes #193

@runame please review and feel free to recommend the wording about the class-dim caveat.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -19,7 +19,7 @@ jobs:
       # You can test your matrix by printing the current Python version
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade pip wheel packaging
           pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
           pip install -e .
       - name: Test with pytest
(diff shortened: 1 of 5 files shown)


runame (approved):
LGTM

PR #216, opened 2024-08-05
Fix ` FullSubnetLaplace.sample()`

### LINKED ISSUE #213: `FullSubnetLaplace.sample()` doesn't use `generator`
Hi

I was doing some reproducibility tests for `pred_type="nn"`, so testing `SomeLaplaceClass._nn_predictive_samples()` -> `SomeLaplaceClass.sample()`, using the current `main` (508843dbc37498e02ef421a573e7c546a90a9443).

I found that `FullSubnetLaplace.sample()` doesn't use the passed `generator` (`ruff check --select ARG laplace/subnetlaplace.py` also catches this). The reason is probably that the used `torch.distributions.MultivariateNormal` doesn't offer an API to pass in a `generator`.

I guess it would not be too difficult to rewrite this in the way other sample methods do things -- generate and re-use a set of standard Normal samples (because of #91) and go from there.

Thanks.

elcorto:
Closes #213

This fix is based on inspecting the base class code `FullLaplace.sample()` and `DiagSubnetLaplace.sample()`. It seems to produce sensible results but a second pair of eyes is needed here. I also added a test, since the tests so far did not catch the reported issue.

### DIFF
--- a/laplace/subnetlaplace.py
+++ b/laplace/subnetlaplace.py
@@ -192,9 +192,10 @@ class FullSubnetLaplace(SubnetLaplace, FullLaplace):
     def sample(
         self, n_samples: int = 100, generator: torch.Generator | None = None
     ) -> torch.Tensor:
-        # sample only subnetwork parameters and set all other parameters to their MAP estimates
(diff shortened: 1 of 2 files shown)


wiseodd (approved):
Clean PR as always! LGTM!

PR #218, opened 2024-08-12
Document pros and cons of subnet selection methods

### LINKED ISSUE #217: Questions about defining a subnet by disabling gradients
I have some questions regarding the method of defining a subnetwork by disabling grads, termed here "subnet-by-disable-grads". 
Please excuse the long-ish text below. If there is another place where questions rather than bug reports should be discussed, please let me know (e.g. github discussions).

From #213:

> I have to admit that we don't pay too much attention to the `SubnetLaplace` implementation anymore since the same can be done more intuitively by switching off the unwanted subset's grads. The benefit of the latter is that the Jacobian for the linearized predictive will only be computed over the selected subset. Meanwhile, the former computes the _full_ Jacobian first and then slices it using the subnet indices.

I guess you are referring to [this part of the docs](https://aleximmer.github.io/Laplace/#subnetwork-laplace_1) and [these tests](https://github.com/aleximmer/Laplace/blob/508843dbc37498e02ef421a573e7c546a90a9443/tests/test_subset_params.py). I assumed that this is just another way of selecting a subset of weights, but that is not the case as it seems.

When using `SubnetLaplace`, I do see a speedup when doing `la.fit(...)` and in hyperparameter optimization. So I assume the calculation of the full Jacobian you refer to happens during calls to `_glm_predictive_distribution()`, where one won't save compute and memory?

Is this also true for `subset_of_weights="last_layer"`, i.e. `LLLaplace`? I guess not since the code looks as if it cuts out the last layer first and then operates only on that. If so, then what would be the benefit of using "subnet-by-disable-grads" over `LLLaplace`?

There is `laplace.utils.subnetmask.LastLayerSubnetMask`, but I guess that is only for testing things given that there are two other methods for defining a last-layer subnet (`LLLaplace` and subnet-by-disable-grads).

The examples for "subnet-by-disable-grads" I have seen so far seem to focus on last layer, or more general disabling grads layer-wise. Since one cannot use any of the `Largest*SubnetMask` or `RandomSubnetMask` to disable individual weight's grads in a tensor via a `parameters_to_vector()` + `vector_to_parameters()` round trip, the method seems to be limited to doing `last_layer`-type subnet selection. Is this correct?

The test in #216 checks that in case of using `SubnetLaplace`, only non-fixed parameters vary across samples. I think this behavior is unique to `SubnetLaplace`-based selections, since the `sample()` method of `LLLaplace` and its subclasses return only samples of the last layer weights, which is corrected for in `_nn_predictive_samples()` as far as I can see. I wonder if `sample()` in case of subnet-by-disable-grads is aware of disabled grads, since all methods which are not in subclasses of `SubnetLaplace` or `LLLaplace` seem to generate "just" vectors of length `n_params` such that parameter samples would also vary for fixed parameters.

Thanks.

elcorto:
Closes #217

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,9 @@ a subnetwork within a neural network (while keeping all other parameters
 fixed at their MAP estimates), as proposed in [11]. It also exemplifies
 different ways to specify the subnetwork to perform inference over.
 
+First, we make use of `SubnetLaplace`, where we specify the subnetwork by
+generating a list of indices for the active model parameters.
+
 ```python
 from laplace import Laplace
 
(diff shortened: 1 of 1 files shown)


wiseodd (approved):
LGTM - thanks a lot!

PR #224, opened 2024-08-19
Pin `asdl` and vendor previous version in `laplace.third_party`

kourbou:
This PR fixes pip installation on `main` by pinning the version of `asdl` to commit [`011a942`](https://github.com/kazukiosawa/asdl@011a942) and vendoring the previous version in `laplace.third_party`. I've copied over the LICENSE file from `asdl` but feel free to let me know how else I can give proper attribution. Maybe we should consider linking the original source in a doc-string in `__init__.py`.


aleximmer:
Thanks for preparing this draft. If we go down this route, I would suggest removing all functionality that we do not need, for example `fr.py` and `hessian.py` and `precondition.py` could be such examples. Should be easy to check for dependencies automatically and keep what's required. Or is there some argument for keeping all old functionality as well?

kourbou:
Yes, we can remove unused submodules. Test coverage for the old backend should be good enough for keeping only relevant code. I think the argument in favor of keeping all the code is if users of laplace-torch want to re-use other components of ASDL in a same project, but then it's better for them to migrate to the new backend anyways?

wiseodd:
We decided to go with #225. Thanks a lot, nevertheless, @kourbou!

PR #225, opened 2024-08-19
Make importing `asdfghjkl` conditional to the installed optional dependency

wiseodd:
@runame @aleximmer please review.

This should make `pip install laplace-torch` work with or without the old `asdfghjkl` dependency. To try this:

```
pip install -e .
pytest

pip install git+https://git@github.com/wiseodd/asdl@asdfghjkl
pytest
```

@runame if everything is good in this PR, you can directly upload it on PyPi.

Also, if you know a more elegant solution, please let me know.


wiseodd:
Added both cases above to the CI.

aleximmer (approved):

PR #231, opened 2024-08-21
Fix typo in the default value of `dict_key_x`

wiseodd:
Change the default from `inputs_id` to `input_ids`, which is the default of Huggingface.

Based on https://github.com/wiseodd/llm-bayesopt-exps/issues/2


aleximmer (approved):
Is there any changes to the docs  that has to be reviewed or is it just because of some version change?

runame (approved):
Same question as Alex's.

wiseodd:
Thanks! The docs are fine --- I double-checked it.

Issue #203, opened 2024-07-04
Efficient, universal, standalone Jacobian backend


wiseodd:
- Functorch = memory blowup due to `vmap`
- Asdl/asdfghjkl = can't backprop through the Jacobians => can't be used for continuous BO
- BackPACK = requires inflexible extension

We need a Jacobian backend that is:
- memory efficient (think about LLM applications!)
- backpropable
- applicable to general type of models (like functorch)

aleximmer:
In my experience, asdl works well for differentiable Jacobian computation. However, I would hope that curvlinops eventually offers such functionality as well since asdl is not actively maintained and, specifically, the differentiability aspect is only available in branches that have not been merged. It also is not present anymore in asdfghjkl, where differentiability broke for some reason I couldn't figure out.

wiseodd:
### Discussion result with @f-dangel

Let `J(x)SJ(x)^T: \R^k \to R^k`, i.e. `(J(x)SJ(x)^T)(v)` for `v \in \R^k`.

#### Computation with vmap (for small `K` or want differentiability)

Then `vmap(J(x)SJ(x)^T)(I)` gives us the `var(f(x))`. Note that this is `vmap` over num classes, unlike the current Jacobian implementation.

```python
def model_fn_params_only(params_dict, buffers_dict):
      out = torch.func.functional_call(self.model, (params_dict, buffers_dict), x)
      return out, out

_, vjpfunc = torch.func.vjp(model_fn_params_only, self.params_dict, has_aux=True)
_, jvpfunc = torch.func.jvp(model_fn_params_only, self.params_dict, has_aux=True)

def JSJT(v):
    v = vjpfunc(v)
    v = S @ v
    v = jvpfunc(v)
    return v

func_var = vmap(JSJT)(I)
```

#### Computation with for-loop (for large K & don't need differentiability)

- `stack`
  - `(k, k)`
  - More memory efficient than computing `J(x)SJ(x)^T` explicitly
    from `J(x)` and `S` since we only store `(k, 1)` or `(p, 1)` tensor each time.

```python
func_var = torch.stack([JSJT(v).detach() for v in I])
```

If only care about diag of `func_var`:

```python
func_var = torch.stack([JSJT(v).detach()[i] for (i, v) in enumerate(I)])
```

#### Sampling

For sampling `f(x)` this can be done cheaply (see Laurence paper).
For LLMs, this might be better because we don't really care about the explicit `J(x)SJ(x)`,
[...]

PR #238, opened 2024-09-01
Integrate previous version of `asdfhjkl` directly into Laplace

aleximmer:
First version integrating the old asdfghjkl into Laplace directly. This allows to have it alongside the latest asdl and further integrate existing extensions, such as end-to-end differentiability and support for other loss functions. This version does not change any behavior and only integrates functions of asdfghjkl that are actually used. The only exception to this is `kernel.py`, which is currently not yet used but will be worth integrating. 

Points to discuss:
- What about documentation for `asdfghjkl_src`?
- Should we first make sure `asdfghjkl` can be default by enabling regression?
- If  `asdfghjkl` becomes default, what is a sensible way to integrate it? I think we could have `curvature/asd` for the core utilities and merge the interfaces/default backends in `curvature.py` with `asdfghjkl.py`, deprecate the `AsdfghjklXYZ` classes and just go with `GGN, EF, Hessian` for them.


wiseodd:
Thanks for the progress!

* **Documentation:** Since we are going to integrate it into `laplace`'s core, we should follow `laplace`'s code standard, in particular, tests, typehinting and docstrings. I think the last two are enough for documentation.
* **Minimum required features:** I vote to make it default since all the backends currently have one limitations and the others (https://github.com/aleximmer/Laplace/issues/203). To do so, I believe these are the requirements:
  - [ ] Support for `regression`
  - [ ] Support for backdrop through the Jacobian & GLM predictive (for continuous Bayesian optimization)
  - [ ] Support for LoRA + Huggingface models + reward modeling. The keys here are: (I think these are automatically supported, at least this is the case for the new `ASDL`. In any case, we need to provide tests for these.)
     - [ ] Support for arbitrary inputs (`x: Union[Tensor, MutableMapping[str, Any]]`).
     - [ ] Support for subset-of-weights (the Hessian & Jacobians are computed only for parameters with `requires.grad == True`).
* **Integration:** I agree with you. We can put the forked `asdfghjkl` code in something like `laplace.curvature.default_backend` and use it directly in `laplace.curvature.CurvatureInterface` to provide a default implementations. So, if another backend doesn't implement a functionality, it will be covered by this default backend. Note that currently this default functionality is provided by `torch.func` which is not efficient.

aleximmer:
Thanks, sounds good to me. I think `asdfghjkl` is well suited to be the universal backend we need and is much faster than `torch.func` for Jacobians etc. I will take care of the points you mentioned.

Issue #232, opened 2024-08-22
OOM when try to fit LA on ViT


ruili-pml:
Hi,

I'm trying to do a subnetwork laplace with diagonal covariance on a ViT base. For now I'm trying to make the mlp part in the transformer block Bayesian and I got OOM error, which is caused by the Jacobian computation. As I'm not familiar with the backend computation of Jacobian, I was wondering is this something I can fix relatively easily or it's just not feasible?

Thanks.
Rui

` File "/scratch/work/src/vit_la_fit.py", line 51, in <module>
    la.fit(train_loader)
  File "/scratch/work/lir3/Laplace/laplace/baselaplace.py", line 703, in fit
    loss_batch, H_batch = self._curv_closure(X, y, N=N)
  File "/scratch/work/lir3/Laplace/laplace/baselaplace.py", line 1758, in _curv_closure
    return self.backend.diag(X, y, N=N, **self._asdl_fisher_kwargs)
  File "/scratch/work/lir3/Laplace/laplace/curvature/curvature.py", line 417, in diag
    Js, f = self.last_layer_jacobians(x) if self.last_layer else self.jacobians(x)
  File "/scratch/work/lir3/Laplace/laplace/curvature/curvature.py", line 115, in jacobians
    Js, f = torch.func.jacrev(model_fn_params_only, has_aux=True)(
  File "/scratch/work/lir3/.conda_envs/dbnn_env/lib/python3.9/site-packages/torch/_functorch/eager_transforms.py", line 601, in wrapper_fn
    flat_jacobians_per_input = compute_jacobian_stacked()
  File "/scratch/work/lir3/.conda_envs/dbnn_env/lib/python3.9/site-packages/torch/_functorch/eager_transforms.py", line 532, in compute_jacobian_stacked
    chunked_result = vmap(vjp_fn)(basis)
  File "/scratch/work/lir3/.conda_envs/dbnn_env/lib/python3.9/site-packages/torch/_functorch/apis.py", line 188, in wrapped
    return vmap_impl(func, in_dims, out_dims, randomness, chunk_size, *args, **kwargs)
  File "/scratch/work/lir3/.conda_envs/dbnn_env/lib/python3.9/site-packages/torch/_functorch/vmap.py", line 266, in vmap_impl
    return _flat_vmap(
  File "/scratch/work/lir3/.conda_envs/dbnn_env/lib/python3.9/site-packages/torch/_functorch/vmap.py", line 38, in fn
[...]

PR #239, opened 2024-09-02
Fix `SubnetLaplace` backend check

wiseodd:
Based on #232

### DIFF
--- a/laplace/subnetlaplace.py
+++ b/laplace/subnetlaplace.py
@@ -64,7 +64,7 @@ class SubnetLaplace(ParametricLaplace):
     temperature : float, default=1
         temperature of the likelihood; lower temperature leads to more
         concentrated posterior and vice versa.
-    backend : subclasses of `laplace.curvature.CurvatureInterface`
+    backend : subclasses of `laplace.curvature.{GGNInterface,EFInterface}`
         backend for access to curvature/Hessian approximations
     backend_kwargs : dict, default=None
         arguments passed to the backend on initialization, for example to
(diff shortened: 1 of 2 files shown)


laplace/subnetlaplace.py
```diff
@@ -99,11 +99,8 @@ def __init__(
             backend_kwargs=backend_kwargs,
         )
 
-        if backend is not None:
-            if not isinstance(backend, GGNInterface) and not isinstance(
-                backend, EFInterface
-            ):
-                raise ValueError("SubnetLaplace can only be used with GGN and EF.")
+        if backend is not None and not issubclass(backend, (GGNInterface, EFInterface)):
```
aleximmer, line 102: Use `isinstance` instead so the vanilla GGN/EF interfaces can be used or are they considered subclasses of themselves?

tests/test_subnetlaplace.py
```diff
@@ -32,6 +36,17 @@
...
+    AsdlEF,
+    AsdlGGN,
+    AsdlHessian,
+    BackPackEF,
+    BackPackGGN,
+    AsdlHessian,
+    CurvlinopsEF,
+    CurvlinopsGGN,
+    CurvlinopsHessian,
+]
```
aleximmer, line 49: Add vanilla backend?

aleximmer (approved):

laplace/subnetlaplace.py
wiseodd: The latter is already true: https://docs.python.org/3/library/functions.html#issubclass

aleximmer:
I think you can merge unless @runame wants to have a look as well.

runame:
LGTM

runame (approved):
LGTM

PR #240, opened 2024-09-02
Transform `y` into 2D tensor when `y.ndim == 1` and `likelihood == REGRESSION`

### LINKED ISSUE #235: Confusing behavior for one-dimensional targets
I noticed that for one-dimensional targets the library produces erroneous results if the targets miss a target dimensional (i.e. only have a batch dimension). Here is a quick example with a linear model for which the Hessian is straightforward to compute:

```import torch
from torch.utils.data import TensorDataset, DataLoader
import laplace

N=10 # number of datapoints
in_d=5 # input dimension

# generate dataset
X=torch.randn((N,in_d))
Y=torch.randn((N,)) # no target dimension
data = TensorDataset(X,Y)
dataloader = DataLoader(data, batch_size=5)

# linear net
model = torch.nn.Linear(in_d,1,bias=False)

# Laplace Approximation
la = laplace.FullLaplace(model=model, likelihood='regression')
la.fit(dataloader)

# Compare la.H with theoretical value for Hessian
H_theory = torch.einsum('ij,ik->jk',X,X)
print(torch.all(torch.isclose(la.H, H_theory))) # will yield False
```
Once you modify `Y=torch.randn((N,))` to `Y=torch.randn((N,1))` the correct Hessian is returned. 

While PyTorch itself throws indeed a warning, it's really easy to overlook when lots of other output is produced. An error message from the Laplace Library would have saved me a lot of time.

wiseodd:
Fixes #235 

Rationale on why we don't want to throw an error:

- PyTorch works on both these 1D-output cases:
  - `model(x).squeeze()` with `y.shape == model(x).shape[0:-1]`
  - `model(x)` with `y.shape == model(x).shape`
- So, in the wild, users' dataloaders vary between those to `y.shape`'s. It's too much to ask to require `y.shape == model(x).shape` since PyTorch doesn't.

### DIFF
--- a/laplace/baselaplace.py
+++ b/laplace/baselaplace.py
@@ -847,7 +847,7 @@ class ParametricLaplace(BaseLaplace):
                 X, y = data
                 X, y = X.to(self._device), y.to(self._device)
 
-            if self.likelihood == Likelihood.REGRESSION and y.ndim == 1:
+            if self.likelihood == Likelihood.REGRESSION and y.ndim == out.ndim - 1:
                 y = y.unsqueeze(-1)
 
             self.model.zero_grad()
@@ -2233,10 +2233,10 @@ class FunctionalLaplace(BaseLaplace):
                 X, y = data
                 X, y = X.to(self._device), y.to(self._device)
 
-            if self.likelihood == Likelihood.REGRESSION and y.ndim == 1:
-                y = y.unsqueeze(1)
-
             Js_batch, f_batch = self._jacobians(X, enable_backprop=False)
+
+            if self.likelihood == Likelihood.REGRESSION and y.ndim == f_batch.ndim - 1:
+                y = y.unsqueeze(1)
 
             with torch.no_grad():
                 loss_batch = self.backend.factor * self.backend.lossfunc(f_batch, y)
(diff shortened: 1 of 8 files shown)


laplace/baselaplace.py
```diff
@@ -846,6 +846,10 @@ def fit(
             else:
                 X, y = data
                 X, y = X.to(self._device), y.to(self._device)
+
+            if self.likelihood == Likelihood.REGRESSION and y.ndim == out.ndim - 1:
+                y = y.unsqueeze(-1)
```
aleximmer, line 851: Should we also raise a warning in this case? It might be that the model was trained the wrong way because the standard MSELoss also assumes that `y` and `out` have the same dimensionality.

aleximmer (approved):
Thanks for fixing it. However, I would slightly be in favor of either raising an error or at least a warning if `out.ndim != y.ndim` because that's the same behavior of `MSELoss` afaik?

Also a test is failing, any idea why?

laplace/baselaplace.py
wiseodd: Thanks, I think raising an error is the way to go. Updated!

runame (approved):
LGTM now

PR #247, opened 2024-09-19
Make `*Laplace` respect the `dtype` of the base model

### LINKED ISSUE #246: Library breaks double precision
I noticed that some parts of the library produce float results even when only objects of double precision are being used. Here is a quick example 
```
import torch
import torch.nn as nn
from torch.utils.data import TensorDataset, DataLoader
import laplace
dtype = torch.float64
X = torch.randn((100,3), dtype=dtype)
Y = torch.randn((100,3), dtype=dtype)
data = TensorDataset(X,Y)
dataloader = DataLoader(data, batch_size=10)
model = nn.Linear(3,3, dtype=dtype)
full_la = laplace.Laplace(model=model, subset_of_weights='all',
        likelihood='regression', hessian_structure='full')
full_la.fit(dataloader)
print(full_la.H.dtype) # prints torch.float32 (at least on my machine) 
``` 
Interestingly when using KFAC instead of the full hessian (i.e.  when `full` above is replaced by `kron`) the hessian `la.H.to_matrix` is of dtype `float.64`.

wiseodd:
Closes #246


wiseodd:
This seems to be a larger endeavor than expected: not only `*Laplace` classes but also all util functions and backends need to be checked.

Marking it as a draft for now.

laplace/baselaplace.py
```diff
@@ -319,17 +334,21 @@ def prior_precision(self, prior_precision: float | torch.Tensor):
...
+                self._prior_precision = (
+                    prior_precision.reshape(-1).to(self._device).to(self._dtype)
+                )
             elif prior_precision.ndim == 1:
                 if len(prior_precision) not in [1, self.n_layers, self.n_params]:
                     raise ValueError(
                         "Length of prior precision does not align with architecture."
                     )
-                self._prior_precision = prior_precision.to(self._device)
+                self._prior_precision = prior_precision.to(self._device).to(self._dtype)
```
runame, line 351: You can use `.to(device=self._device, dtype=self._dtype)` instead.

laplace/baselaplace.py
```diff
@@ -547,14 +567,16 @@ def sigma_noise(self, sigma_noise: float | torch.Tensor) -> None:
         self._posterior_scale = None
 
         if np.isscalar(sigma_noise) and np.isreal(sigma_noise):
-            self._sigma_noise = torch.tensor(sigma_noise, device=self._device)
+            self._sigma_noise = torch.tensor(
+                sigma_noise, device=self._device, dtype=self._dtype
```
runame, line 571: same as above, `as_tensor` could be used; applies everywhere `tensor` is used.

tests/test_baselaplace.py
```diff
@@ -855,3 +855,42 @@ def test_parametric_fit_y_shape(model_1d, reg_loader_1d, reg_loader_1d_flat, lap
...
+def test_dtype(laplace, backend, dtype, likelihood):
+    X = torch.randn((10, 3), dtype=dtype)
+    Y = torch.randn((10, 3), dtype=dtype)
+
+    data = TensorDataset(X, Y)
+    dataloader = DataLoader(data, batch_size=10)
+
+    model = nn.Linear(3, 3, dtype=dtype)
+
+    try:
```
runame, line 875: I think this test would pass if there is a `ValueError` because of trying to perform an operation on two tensor with different dtype. Can we make the try-except less general?

tests/test_functional_laplace_unit.py
```diff
@@ -322,3 +326,46 @@ def test_functional_fit_y_shape(model_1d, reg_loader_1d, reg_loader_1d_flat):
...
+def test_dtype(laplace, backend, dtype, independent_outputs, likelihood):
+    X = torch.randn((10, 3), dtype=dtype)
+    Y = torch.randn((10, 3), dtype=dtype)
+
+    data = TensorDataset(X, Y)
+    dataloader = DataLoader(data, batch_size=10)
+
+    model = nn.Linear(3, 3, dtype=dtype)
+
+    try:
```
runame, line 345: same here.

tests/test_lllaplace.py
```diff
@@ -699,3 +703,42 @@ def test_reg_glm_predictive_correct_behavior(laplace, model, reg_loader):
...
+def test_dtype(laplace, backend, dtype, likelihood):
+    X = torch.randn((10, 3), dtype=dtype)
+    Y = torch.randn((10, 3), dtype=dtype)
+
+    data = TensorDataset(X, Y)
+    dataloader = DataLoader(data, batch_size=10)
+
+    model = nn.Linear(3, 3, dtype=dtype)
+
+    try:
```
runame, line 723: same here.

tests/test_subnetlaplace.py
```diff
@@ -914,3 +914,44 @@ def test_sample(model, likelihood, hessian_structure, class_loader, reg_loader):
...
+
+    data = TensorDataset(X, Y)
+    dataloader = DataLoader(data, batch_size=10)
+
+    model = nn.Linear(3, 3, dtype=dtype)
+
+    subnetmask = RandomSubnetMask(model=model, n_params_subnet=10)
+    subnetmask.select()
+
+    try:
```
runame, line 937: same here.

wiseodd:
@aleximmer could you please double check and merge?

aleximmer:
Thanks @wiseodd, LGTM!

PR #261, opened 2024-12-02
Update dev guide re. publishing

wiseodd:
@runame @aleximmer 

I don't have permission on TestPyPI so I can test it there.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,9 +42,6 @@ build-backend = "pdm.backend"
 [tool.setuptools]
 py-modules = ["laplace"]
 
-[tool.pdm]
-distribution = true
-
 [tool.uv]
 default-groups = ["docs"]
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -42,9 +42,6 @@ build-backend = "pdm.backend"
 [tool.setuptools]
 py-modules = ["laplace"]
 
-[tool.pdm]
```
wiseodd, line 45: This is because we don't recommend use `pdm` as tool anymore. Note that the `pdm-backend` build backend is separate from `pdm` itself.

wiseodd: `uv` has a plan to release its own build-backend. We will replace `pdm-backend` when the time comes.

runame (approved):
Not familiar with these `uv` commands, but LGTM.

I just added you as an owner on TestPyPI, so you can test it.

aleximmer (approved):
LGTM. Have you tried this out with the permissions given by @runame ?

wiseodd:
Yup, works: https://test.pypi.org/manage/project/laplace-torch/release/0.2.2.2/

PR #263, opened 2024-12-04
Expectation examples

wiseodd:
Example of how to leverage the newly added `functional_samples` method #243. 

I also added the abstract interface for `samples` and `functional_samples` to `BaseLaplace`.

### DIFF
--- a/docs/expectation_example.md
+++ b/docs/expectation_example.md
@@ -1,0 +1,59 @@
+## Motivation
+
+Often, we want to compute non-analytic expectation \(\mathbb{E}_{p(f(x) \mid \mathcal{D})} [g(f(x))]\) of a function \(g(f(x))\) given a Laplace posterior over functions \(p(f(x) \mid \mathcal{D})\) at input \(x\).
+This naturally arises in decision-making:
+Given a posterior belief about an unknown function \(f\) we would like to compute the expected utility of \(x\).
+In Bayesian optimization, this is called _acquisition function_.
+For some utility function \(g\) and posterior belief \(p(f(x) \mid \mathcal{D})\), the resulting acquisition function can be computed analytically, e.g. if \(g\) is linear and the posterior is Gaussian (process).
+But, in general, closed-form solutions don't exist. 
+
+In this example, we will see how easy it is to compute a _differentiable_, Monte-Carlo approximated acquisition function under the posterior distribution over neural network functions implied by a Laplace approximation.
+
+## Laplace approximations
+
+As always, the first step of a Laplace approximation is MAP estimation.
+
+```python
+import torch
+import torch.utils.data as data_utils
+from torch import autograd, nn, optim
+
+from laplace import Laplace
+from laplace.utils.enums import HessianStructure, Likelihood, PredType, SubsetOfWeights
+
+torch.manual_seed(123)
+
+model = nn.Sequential(nn.Linear(2, 10), nn.GELU(), nn.Linear(10, 1))
+X, Y = torch.randn(5, 2), torch.randn(5, 1)
+train_loader = data_utils.DataLoader(data_utils.TensorDataset(X, Y), batch_size=3)
+opt = optim.AdamW(model.parameters(), lr=1e-3, weight_decay=5e-4)
+loss_fn = nn.MSELoss()
+
+for epoch in range(10):
+    model.train()
+
+    for x, y in train_loader:
+        opt.zero_grad()
+        out = model(x)
+        loss = loss_fn(out, y)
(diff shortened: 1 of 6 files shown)


laplace/baselaplace.py
```diff
@@ -694,6 +694,62 @@ def _glm_forward_call(
                 "Prediction path invalid. Check the likelihood, pred_type, link_approx combination!"
             )
 
+    def sample(
```
runame, line 697: Random thought and I probably don't remember enough context, but it seems like it would make sense for `BaseLaplace` to inherit from `abc.ABC` and use `@abc.abstractmethod` for cases like this one.

examples/expectation_example.py
```diff
@@ -0,0 +1,74 @@
...
+    print(f"Thompson sampling, {pred_type}")
+
+    x_test = torch.randn(10, 2)
+    x_test.requires_grad = True
+
+    f_sample = la.functional_samples(x_test, pred_type=pred_type, n_samples=1)
+    f_sample = f_sample.squeeze(0)  # We only use a single sample
+    print(f"TS shape: {f_sample.shape}, TS requires grad: {f_sample.requires_grad}")
+
+    grad_x = autograd.grad(f_sample.sum(), x_test)[0]
```
runame, line 47: A comment describing what is happening here would be nice.

examples/expectation_example.py
```diff
@@ -0,0 +1,74 @@
...
+
+    print(
+        f"Grad x_test shape: {grad_x.shape}, Grad x_test vanishing: {torch.allclose(grad_x, torch.tensor(0.0))}"
+    )
+    print()
+
+print()
+
+# Monte-Carlo expected improvement E_{f(x) ~ p(f(x) | D)} [max(f(x) - best_f, 0)]
```
runame, line 56: Introduce later used abbreviation `(IE)` here.

examples/expectation_example.py
```diff
@@ -0,0 +1,74 @@
...
+    x_test = torch.randn(10, 2)
+    x_test.requires_grad = True
+
+    f_samples = la.functional_samples(x_test, pred_type=pred_type, n_samples=10)
+    ei = (f_samples - f_best.reshape(1, 1, 1)).clamp(0.0).mean(0)
+    print(f"EI shape: {ei.shape}, EI requires grad: {ei.requires_grad}")
+
+    grad_x = autograd.grad(ei.sum(), x_test)[0]
+
+    print(
```
runame, line 71: Same here, adding a few comments describing what you are trying to show would be nice.

docs/expectation_example.md
```diff
@@ -0,0 +1,120 @@
+## Motivation
```
runame, line 1: I have looked at this file after `expectation_example.py`, maybe it is worth it to just copy some of the commentary here to the runnable example (as comments)?

I haven't thought about it much, but there should be a nicer solution than duplicating the example code in a markdown file for the docs and a runnable Python file. Maybe a Jupyter notebook that can be rendered as a markdown file?

runame (approved):
Besides my comments, LGTM.

docs/expectation_example.md
aleximmer: I agree the duplicated code is not great but I think it's already a bit longer like this. We should add an issue to write a small script that extracts python blocks from the md doc files and runs that. I think we introduced having them both as py and md files so we can make sure the examples actually run. Will add an issue for this.

laplace/baselaplace.py
aleximmer: Agree, I think we use the old "implicit" way everywhere in the code at the moment though. So this should be done separately across the whole lib.

uv.lock
```diff
@@ -601,7 +601,7 @@ wheels = [
 
 [[package]]
 name = "laplace-torch"
-version = "0.2.2"
+version = "0.2.2.2"
```
aleximmer, line 604: Why add another separator instead of going to `.3` here?

aleximmer (approved):
LGTM, thanks for adding the example. Some comment on a higher level but not necessary to add: it would be cool to plot the EI function and then compute the gradient at some point and add that to the plot as well to visualize the example mot intuitively.

laplace/baselaplace.py
wiseodd: Let's do this in a new PR https://github.com/aleximmer/Laplace/issues/275

docs/expectation_example.md
wiseodd: https://github.com/aleximmer/Laplace/issues/276

PR #264, opened 2024-12-05
Remove unknown exit() from utils.py

BlackHC:


wiseodd:
Thanks a lot, @BlackHC! LGTM, but could you add a test for this here? https://github.com/aleximmer/Laplace/blob/main/tests/test_utils.py

laplace/utils/utils.py
```diff
@@ -219,7 +219,7 @@ def symeig(M: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
             stats = f"diag: {M.diagonal()}, max: {M.abs().max()}, "
             stats = stats + f"min: {M.abs().min()}, mean: {M.abs().mean()}"
             logging.info(f"SYMEIG: adding jitter failed. Stats: {stats}")
-            exit()
+            raise
```
aleximmer, line 222: Thanks for replacing this silent exit. Could you make it verbose with `raise RuntimeError(f"Adding jitter failed. Stats: {stats}") from e`?

aleximmer (approved):
LGTM, I don't think this necessarily requires a test. A more verbose exception is easy to do though so I would add that.

PR #266, opened 2024-12-06
Support for multidim outputs

### LINKED ISSUE #252: 3D input tensors and feature reduction
@wiseodd 
Tldr; Issue with tensors of size (B, L, D) passing through a `Linear` last layer 

Here's the minimal reproducible example:
```python
import torch
import torch.nn as nn
from laplace import Laplace
from torch.utils.data import DataLoader
from tensordict import TensorDict

BATCH_SIZE = 4  # B
SEQ_LENGTH = 6  # L
EMBED_DIM = 8  # D
INPUT_KEY = "input"
OUTPUT_KEY = "output"


class Model(nn.Module):
    def __init__(self):
        super().__init__()
        self.attn = nn.MultiheadAttention(EMBED_DIM, num_heads=1)
        self.final_layer = nn.Linear(EMBED_DIM, 1)

    def forward(self, x):
        x = x[INPUT_KEY].view(-1, SEQ_LENGTH, EMBED_DIM)  # (B, L, D) 
        out = self.attn(x, x, x, need_weights=False)[0]  # (B, L, D)
        return self.final_layer(out).squeeze(dim=-1)  # (B, L)


ds = TensorDict({INPUT_KEY: torch.randn((100, SEQ_LENGTH * EMBED_DIM)),
                 OUTPUT_KEY: torch.randn((100, SEQ_LENGTH * 1))},
                batch_size=[100])  # simulates a dataset
dl = DataLoader(ds, batch_size=BATCH_SIZE, shuffle=False, collate_fn=lambda x: x)

model = Model()
la = Laplace(model, "regression", dict_key_x=INPUT_KEY, dict_key_y=OUTPUT_KEY,
             last_layer_name="final_layer", feature_reduction="average")
la.fit(dl)

data = next(iter(dl))  # data[INPUT_KEY].shape = (B, L * D)
pred_map = model(data)  # (B, D)
pred_lap = la(data)  # TODO: error! (shape '[4, 6, 54]' is invalid for input of size 216)
```

My goal is to obtain some measure of epistemic uncertainty on each "token" of the output. The difference from the tutorials I reviewed is the 3D input tensor, which I need for attention. Using the `feature_reduction` parameter seems to help a little when I was trying to debug but I'm not very familiar with this functionality.

I find it surprising that `la.fit(dl)` works but the forward call `la(data)` doesn't work. How do you recommend I use this library properly for this use-case? Thanks in advance.

wiseodd:
Closes #252 

Thanks to Kazuki for updating `asdfghjkl` on PyPI, we can now handle 3D tensor outputs (e.g. in LLMs) with GLM predictive.

### DIFF
--- a/laplace/baselaplace.py
+++ b/laplace/baselaplace.py
@@ -158,7 +158,7 @@ class BaseLaplace:
         )
 
         # log likelihood = g(loss)
-        self.loss: float = 0.0
+        self.loss: float | torch.Tensor = 0.0
         self.n_outputs: int = 0
         self.n_data: int = 0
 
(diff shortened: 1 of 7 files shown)


wiseodd:
WIP adding an article based on `lm_example.py` to the docs.

wiseodd:
@runame @aleximmer ready to review!

laplace/baselaplace.py
```diff
@@ -256,15 +315,16 @@ def predictive(
...
+        )
 
     def _check_jacobians(self, Js: torch.Tensor) -> None:
         if not isinstance(Js, torch.Tensor):
             raise ValueError("Jacobians have to be torch.Tensor.")
         if not Js.device == self._device:
             raise ValueError("Jacobians need to be on the same device as Laplace.")
-        m, k, p = Js.size()
-        if p != self.n_params:
+        if Js.shape[-1] != self.n_params:
```
aleximmer, line 327: For consistency with previous assertion, perhaps add `and Js.ndim >= 3`?

laplace/baselaplace.py
```diff
@@ -1335,7 +1397,7 @@ def _nn_predictive_classification(
 
         vector_to_parameters(self.mean, self.params)
 
-        return py
+        return torch.as_tensor(py)
```
aleximmer, line 1400: Just to make sure: I suppose this is for typing?

laplace/curvature/asdl.py
```diff
@@ -83,16 +83,23 @@ def closure():
                 )
                 return f
 
-            Ji, f = batch_gradient(
-                self.model,
-                closure,
-                return_outputs=True,
-                batch_size=self._get_batch_size(x),
-            )
+            Ji, f = batch_gradient(self.model, closure, return_outputs=True)
```
aleximmer, line 86: Why is the `batch_size` argument no needed anymore, or perhaps, why was it even needed before?

tests/test_baselaplace.py
aleximmer, line 1: Could you add a test where you test for equivalence between a "trivial" multidimensional model and a previously supported model? This can be just a linear model that mapped previously from `[num_data, D] -> [num_data, K]` and now from `[num_data / L, L, D] -> [num_data / L, L, K]`, that is, the data set is simply reshaped. This should give an equivalent posterior and predictive. Maybe to make sure also have the option to use a NN for the test, not only the linear model case.

aleximmer (changes requested):
Thanks for the useful addition. To make sure this breaks nothing, I would recommend the test mentioned in the review, otherwise LGTM.

docs/lm_example.md
```diff
@@ -0,0 +1,139 @@
...
+
+```python
+ds = TensorDict(
+    {
+        INPUT_KEY: torch.randn((100, SEQ_LENGTH, EMBED_DIM)),
+        OUTPUT_KEY: torch.randn((100, SEQ_LENGTH, OUTPUT_SIZE)),
+    },
+    batch_size=[100],
+)  # simulates a dataset
+dl = DataLoader(ds, batch_size=BATCH_SIZE, shuffle=False, collate_fn=lambda x: x)
```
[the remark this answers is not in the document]
wiseodd: Yes, it's required for this example. FYI though, Huggingface also uses non-default `collate_fn`

laplace/baselaplace.py
wiseodd: Yes!

laplace/curvature/asdl.py
wiseodd, line 86: @runame @aleximmer do you know what happened here?

```python
...
Ji, f = batch_gradient(self.model, closure, return_outputs=True)
print("Jac")
print(Ji)
...
```

with these setup: 

```python
X, Y = torch.randn(10, 3), torch.randn(10, 1)
X_multidim, Y_multidim = X.reshape(5, 2, 3), Y.reshape(5, 2, 1)

model_std = nn.Linear(3, 1)
model_multidim = deepcopy(model_std)

...

la = Laplace(model_multidim, ...)
la.fit(...)

la(X_multidim)
```

prints

```
Jac
tensor([[ 0.2302,  0.1006,  0.4197,  1.0000],
        [ 0.0000,  0.0000,  0.0000,  0.0000],
        [ 0.3055,  1.1884, -0.4010,  1.0000],
        [ 0.0000,  0.0000,  0.0000,  0.0000],
        [-2.0167,  1.4216, -0.5488,  1.0000],
        [ 0.0000,  0.0000,  0.0000,  0.0000],
        [-0.4076,  0.2810, -1.7831,  1.0000],
        [ 0.0000,  0.0000,  0.0000,  0.0000],
        [ 0.1297,  0.2202, -0.5024,  1.0000],
        [ 0.0000,  0.0000,  0.0000,  0.0000]]
```

wiseodd: Notice that the returned `Ji` has row `[0 ... 0]`. Seems like `ASDL`'s issue?

wiseodd: To reproduce: `uv run pytest -k "test_predictive_multidim`

PR #282, opened 2025-09-10
Fix subnet samples (`posterior_scale.T` issue)

### LINKED ISSUE #274: Potential bug in `FullSubnetLaplace.sample()`
Looking at #268, I think the same issue is present in [`FullSubnetLaplace.sample()`](https://github.com/aleximmer/Laplace/blob/33d7906cfabbcf12c95ab85affb98304297fe232/laplace/subnetlaplace.py#L203) since it uses the same construct `samples @ self.posterior_scale`, which was inspired by `FullLaplace.sample` in #216. It might also affect other `sample` methods, such as [the one in `DiagSubnetLaplace`](https://github.com/aleximmer/Laplace/blob/33d7906cfabbcf12c95ab85affb98304297fe232/laplace/subnetlaplace.py#L239), but I'd need to inspect that code more closely. 

In any case I think before fixing this, one needs tests catching the issue. My first guess is to draw (potentially many) samples, calculate the empirical cov matrix and compare against `posterior_covariance`, as @2bys and @simoneiriksson have suggested.

elcorto:
Closes #274 

This fixes the `posterior_scale.T` issue discusssed in #274 (`FullSubnetLaplace.sample()`) and adds tests to make sure `FullSubnetLaplace.sample()` and `DiagSubnetLaplace.sample()` do the right thing. 

We also add cov sampling tests for all other Laplace variants in `tests/test_baselaplace.py` and `tests/test_lllaplace.py`.

### DIFF
--- a/laplace/subnetlaplace.py
+++ b/laplace/subnetlaplace.py
@@ -200,7 +200,7 @@ class FullSubnetLaplace(SubnetLaplace, FullLaplace):
             dtype=self._dtype,
             generator=generator,
         )
-        subnet_samples = self.mean_subnet[None, ...] + samples @ self.posterior_scale
+        subnet_samples = self.mean_subnet[None, ...] + samples @ self.posterior_scale.T
         return self.assemble_full_samples(subnet_samples)
 
 
--- a/tests/test_baselaplace.py
+++ b/tests/test_baselaplace.py
@@ -384,16 +384,6 @@ def test_laplace_functionality(laplace, lh, model, reg_loader, class_loader):
     lml = lml + 1 / 2 * (prior_prec.logdet() - log_det_post_prec)
     assert torch.allclose(lml, lap.log_marginal_likelihood())
 
-    # test sampling
-    torch.manual_seed(61)
-    samples = lap.sample(n_samples=1)
-    assert samples.shape == torch.Size([1, len(theta)])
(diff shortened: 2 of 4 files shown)


elcorto:
I can reproduce the `pytest-with-asdfghjkl ` fail locally when I install `git+https://git@github.com/wiseodd/asdl@asdfghjkl` as the action does. Then the max error is ~0.7 instead of the allowed 0.01:

```
tests/test_baselaplace.py::test_laplace_functionality[LowRankLaplace-classification] >>>>> tensor(0.7377)
FAILED
tests/test_baselaplace.py::test_laplace_functionality[LowRankLaplace-regression] >>>>> tensor(0.7272)
FAILED
```

but only the combo `LowRankLaplace` + the above `asdfghjkl` lib. Maybe one can xfail that by testing for LowRankLaplace and a specific version (range) of the library? At least that's a quick solution to make the (PR) tests pass. Solving the `asdfghjkl` issue is a separate topic I guess.

elcorto:
The lint fail

```
93 |     assert res != math.nan
   |                   ^^^^^^^^
94 |     assert isinstance(res, float)
95 |     assert res > 0
   |

PLW0177 Comparing against a NaN value; use `math.isnan` instead
```

is not due to changes in this PR. Also, I'm with `ruff` here :)

PR #286, opened 2026-02-28
Change curvlinops dependency version

### LINKED ISSUE #285: Laplace Install breaks with Curvlinops 3.0.0 version
We are using laplace as a dependency in [lightning-uq-box]() and noticed the CI tests failing. I believe the reason is the new `curvlinops==3.0.0` refactored package is breaking imports inside laplace, and in your pyproject.toml file https://github.com/aleximmer/Laplace/blob/44ea62ac1f8cfd6f797f31c75e2cf34414fdddae/pyproject.toml#L26 the curvlinops should perhaps be fixed to 2.0.0, to alleviate.

This can also be reproduced with:

```bash
pip install laplace-torch
python
Python 3.10.14 (main, Mar 21 2024, 16:24:04) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import laplace
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/lib/python3.10/site-packages/laplace/__init__.py", line 10, in <module>
    from laplace.baselaplace import (
  File "/lib/python3.10/site-packages/laplace/baselaplace.py", line 22, in <module>
    from laplace.curvature.curvlinops import CurvlinopsEF, CurvlinopsGGN
  File "/lib/python3.10/site-packages/laplace/curvature/curvlinops.py", line 15, in <module>
    from curvlinops._base import _LinearOperator
ModuleNotFoundError: No module named 'curvlinops._base'
```

sizemore0125:
Fixes #285 

Changed curvlinops dependency version to avoid repo breaking due to recent curvlinops update.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,13 +23,14 @@ requires-python = ">=3.9"
 dependencies = [
   "asdfghjkl == 0.1a4",
   "backpack-for-pytorch",
-  "curvlinops-for-pytorch >= 2.0",
+  "curvlinops-for-pytorch == 2.0.1",
   "numpy",
   "opt_einsum",
   "torch >= 2.0",
   "torchmetrics",
-  "torchvision >= 0.15",
+  "torchvision >= 0.15"
 ]
+
 
 [project.urls]
 Homepage = "https://github.com/aleximmer/Laplace"


aleximmer (approved):
Thanks a lot!

elcorto:
Just hit the same issue. Maybe `"curvlinops-for-pytorch >= 2.0, < 3"` is more idiomatic?

ju-kreber:
Hi, I just hit the same issue #285 . However I would strongly recommend against this version pinning, since this has secondary effects, e.g. forces numpy<2.0.0 from curvlinops-for-pytorch < 3.

It appears to me as the reason for this issue, the missing symbol `_LinearOperator` that can no longer be imported  `from curvlinops._base import _LinearOperator`, is only used for type annotations, at least in `curvlinops.py`.
I would much rather therefore recommend to change the code here. It also appears to be not a good idea to import `_Private` symbols from third-party libraries.

EDIT: While this particular import error can indeed be fixed by removing the missing symbol, there seem to be further incompatibilities with newer versions of curvlinops that should be addressed properly.

PR #287, opened 2026-06-12
Fix no longer existing symbol import from curvlinops library

ju-kreber:
Addressing #285 by replacing the no longer existing symbol `_LinearOperatori` in `curvlinops` with `CurvatureLinearOperator`, which is also not a private symbol.
All returned classes are indeed base classes of `CurvatureLinearOperator`.
Importing from `curvlinops.hessian` since currently not exposed in the package root of `curvlinops`.
Refraining from importing from `_torch_base.py` where it is defined.

### DIFF
--- a/laplace/curvature/curvlinops.py
+++ b/laplace/curvature/curvlinops.py
@@ -12,7 +12,7 @@ from curvlinops import (
     HessianLinearOperator,
     KFACLinearOperator,
 )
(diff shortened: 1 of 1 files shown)


ju-kreber:
While the provided pull request does fix the import error, there seem to be several further incompatibilities with never versions of `curvlinops`. Probably they should get addressed at once.

